PR #5, opened 2021-09-23
Extend styling of functional wireframe

mathewjordan:
# What does this do?

Despite the hefty number of commits, this is a pretty basic PR that addresses the initial layout of a functional wireframe. The primary functionality here is a fairly responsive layout with `overflow-x` and `overflow-y` being applied to the body of the Media and Navigator components. This PR does not address any actual state changes such as button toggles or media updates. Some baseline Northwestern styling has been applied though there is much refinement to be completed there.

To pair with this, I also parsed JSON output of a converted sample webVTT file and initial some functionality there. Functionality to fetch and convert that webVtt is not included here. Time parsed from the webVtt as international standard notation (00:01:23.456) is converted to a _clean time_, ex: `00:01:23.456` becomes transforms to `1:23`.

image

- Outline basic viewer components.
- Add IIIF label.
- Add jest, test label output.
- Add Media/Navigator components.
- Rename top level components.
- Cleanup baseline.
- Test drive stiches.
- Apply stiches throughout.
- Add pres 3 video manifest.
- Add sample multi canvas video manifest with vtt tracks.
- Combine player and video components.
- Update basic styling.
- Ignore compiled script.
- Cleanup virtual dom structure. :hammer:
- Fill flex area.
- Mock overflow scroll of Media.
- Add sample media and labels.
- Kickstart active state.
- Apply generic sans font family.
- Refine box visuals.
- Refine font family and padding.
- Make video responsive to window height.
- Update basic virtual dom structure for navigator.
- Refine semantic structure.
- Apply baseline overflow to navigation.
- Style baseline navigator header.
- Update font weights and padding.
- Northwesternize.
- Add gradients. :art:
- Update end of range margin.
- Add baseline webvtt parsing.
- Add helper function to output a pretty time.
- Update cleanTime test.
- Write tests for all digit notations.
- Simplify services directory.


src/components/MediaItem/MediaItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+const sample: object = {
+  label: "Croquet Scene by Winslow Homer",
+  thumbnailId:
+    "https://upload.wikimedia.org/wikipedia/commons/thumb/9/9e/Winslow_Homer_-_Croquet_Scene_-_Google_Art_Project.jpg/640px-Winslow_Homer_-_Croquet_Scene_-_Google_Art_Project.jpg",
+};
+
+const MediaItem: React.FC<Props> = ({ index, label, thumbnailId }) => {
+  return (
+    <MediaItemWrapper onClick={handleUpdate()}>
+      <figure>
```
adamjarling, line 24: Nice to have semantic proper HTML from the get go.

src/components/Navigator/Navigator.tsx
```diff
@@ -1,19 +1,122 @@
 import React, { useEffect, useState } from "react";
 import { styled } from "@stitches/react";
+import NavigatorItem from "components/NavigatorItem/NavigatorItem";
+import sampleParsedVtt from "../../manifests/sampleParsedVtt";
```
adamjarling, line 4: My VSCode is complaining about this line...maybe because of the .json file extension?  We should see if we can match up our TypeScript linting between our IDEs?

src/components/Navigator/Navigator.tsx
```diff
@@ -1,19 +1,122 @@
...
+  return <NavigatorTabWrapper>{label}</NavigatorTabWrapper>;
+};
+
+const NavigatorBodyInner: React.FC<NavigatorBodyInnerProps> = ({
+  items,
+  currentTime,
+}) => {
+  return items.map((cue) => {
+    return (
+      <NavigatorItem
```
adamjarling, line 49: Also getting some TypeScript complaints here...  interesting to know that it will still compile with these complaints...I'll look into that.

adamjarling (approved):
All in all looks great and the main gist of the PR is getting the layout and styling tightened up.  Nice work indeed.  A few small TypeScript things we'll need to iron our for our environment, but can fix that in another PR.

src/components/Navigator/Navigator.tsx
mathewjordan: Yeah, this definitely need some work. I think it compiles because it's a JS map and not a TS map. I was banging my head against apply TS here and just went with what I knew to get it to map without a fail.

src/components/Navigator/Navigator.tsx
mathewjordan: @adamjarling Yeah let's look. Line 4 here won't stay to production, but in the case of importing other JSON files it'd be smart to see how to best handle.

PR #6, opened 2021-09-23
Add local linting

adamjarling:
This adds some local `eslint` config.  Hopefully it's not too big a pain to rebase on top of this branch.


src/components/Navigator/Navigator.tsx
```diff
@@ -40,83 +33,89 @@ const NavigatorTab: React.FC<NavigatorTabProps> = ({ label }) => {
...
+}
+
+const NavigatorBodyInner: React.FC<NavigatorBodyInnerProps> = ({ items, currentTime }) => {
+  return (
+    <>
+      {items.map(({ id, text, startTime }) => (
+        <NavigatorItem label={text} startTime={startTime} t={currentTime} key={id} />
+      ))}
+    </>
+  );
```
mathewjordan, line 55: Wow, this makes so much more sense to me now. I'd like to clean up naming like we discussed but this great. 🧙

src/components/Navigator/Navigator.tsx
```diff
@@ -40,83 +33,89 @@ const NavigatorTab: React.FC<NavigatorTabProps> = ({ label }) => {
...
+const NavigatorWrapper = styled('div', {
+  display: 'flex',
+  flexDirection: 'column',
+  width: '100%',
+  height: '100%',
+  flexGrow: '1',
+  flexShrink: '0',
+  position: 'relative',
+  zIndex: '1',
+  boxShadow: '-5px 0 5px #00000011',
```
mathewjordan, line 67: Just curious @adamjarling, is this something linter was yelling about?

mathewjordan (approved):

src/components/Navigator/Navigator.tsx
adamjarling: If you mean the single quotes, no.  Guess that was an oversight to include that in the prettier config file.  I'm either way on quotes.... and actually am used to Prettier's default way of double quotes.  We can switch it back if ya like.

PR #7, opened 2021-09-23
Manifest fetch and validation using hyperion-framework.

mathewjordan:
The pull request add the hyperion-framework packages **Vault**, **Validator**, **Types**, and **Parser** to our codebase. To test a few of the packages, I just have the label "New Airliner, 1946/02/21" rendering from my hard coded sample manifest.

To test, you'll need to run `npm install` first.

**Vault**
We'll use Vault to (1) fetch a manifest, (2) convert to presentation 3 (if necessary). Vault also stores the manifest using redux (see below) if we want to leverage that. 

image

**Validator**
We also could validate our manifests against the presentation 3 spec using Validator. This has a boolean return  though I have not verified how it determines validation. Validator does not seem to provide verbose reasons why your manifest may be validating. 

**Types**
This is just a helper utility for TS with the types required by the various hyperion packages. It has helpful types for `Manifest`, `Canvas`, `InternationalString`, `MetadataItem`, etc... that we can definitely use going forward.

**Parser**
This is not being used yet, but we'll eventually want to leverage this to pluck and render data delivered by manifests without having to manually traverse things ourselves.

## Notes

image

I am getting a VS Code message on [line 13 of index.tsx](https://github.com/nulib/react-media-player/pull/7/files#diff-0b5adbfe7b36e4ae2f479291e20152e33e940f7f265162d77f40f6bdb5da7405R13) stating:
``` Type '({ id }: PropsWithChildren<Props>) => Element | "A future user friendly loading component"' is not assignable to type 'FC<Props>'.
  Type 'Element | "A future user friendly loading component"' is not assignable to type 'ReactElement<any, any> | null'.
    Type '"A future user friendly loading component"' is not assignable to type 'ReactElement<any, ...```

This is confusing me because I seem to be following the same structure as we are regarding TS and other uses of type of React.FunctionalComponent. Maybe we can chat about this and see what we are doing here that could cause that message. It does not seem to hinder testing or compiling at this time.

### CODE REVIEW

File: src/components/Viewer/Viewer.tsx [hunk 1]
```diff
@@ -1,78 +1,78 @@
...
-          <Player streamingUrl={publicUrl} />
-          <Media items={{}} activeItem={0} />
-        </Main>
-        <Aside>
-          <Navigator currentTime={100} tracks={{}} />
-        </Aside>
-      </ViewerInner>
-    </ViewerWrapper>
-  );
+  if (manifest) {
```
adamjarling, line 19: Totally no big deal, but an alternate way to write this not having to mess with `if else` blocks could be...more Node-ish:

```
if (!imanifest) return null;

return (
   <ViewerWrapper>
  ...
);
```

src/index.tsx
```diff
@@ -1,26 +1,38 @@
...
-    <Viewer
-      manifest={{
-        label: {
-          en: ["manifest.label"],
-        },
-      }}
-    />
-  );
+const App: React.FC<Props> = ({ id }) => {
+  const [manifest, setManifest] = useState<ManifestNormalized | undefined>(undefined);
```
adamjarling, line 14: Do we need the `(undefined)` or can we just leave empty, ie `()`?

src/index.tsx
```diff
@@ -1,26 +1,38 @@
...
+  useEffect(() => {
+    const vault = new Vault();
+    vault
+      .loadManifest(id)
+      .then((data) => {
+        setManifest(data);
+        setLoaded(true);
+      })
+      .catch((error) => {
+        console.log(`Manifest failed to load: ${error}`);
```
adamjarling, line 25: Maybe when throwing errors, we do `console.error()` instead?

src/index.tsx
```diff
@@ -1,26 +1,38 @@
...
-          en: ["manifest.label"],
-        },
-      }}
-    />
-  );
+const App: React.FC<Props> = ({ id }) => {
+  const [manifest, setManifest] = useState<ManifestNormalized | undefined>(undefined);
+  const [loaded, setLoaded] = useState(false);
+  useEffect(() => {
+    const vault = new Vault();
```
adamjarling, line 17: Is it overkill here to put a check in for a truthy `loaded` value?

src/services/iiif.test.ts
```diff
@@ -1,6 +1,12 @@
-import { label } from "./presentation";
+import { getLabel } from './iiif';
```
adamjarling, line 1: Is there a Hyperion helper method to do what `getLabel` is doing here?

adamjarling (commented):
Looks good.  A few random questions/comments but overall nice work!

src/components/Viewer/Viewer.tsx
mathewjordan: @adamjarling Thanks for this, I've never really thought about inverting the logic there.

src/index.tsx
mathewjordan: @adamjarling I'm all for it, can we chat first?

src/services/iiif.test.ts
mathewjordan: @adamjarling There may be through use of Traverse from the parser package. I spent some time looking at this and Traverse seems really helpful for finding **all labels** or **all canvases**. For a top level label it felt like overkill, but I do want to wrap my mind around this more and see where and when we apply Traverse.

adamjarling (approved):
Looks great.

PR #9, opened 2021-09-27
Wrap App with React.Context

mathewjordan:
Related Issues: https://github.com/nulib/repodev_planning_and_docs/issues/2221

## What does this do?

This PR was quite an adventure, however, it introduces work to wrap our app using React.Context and allows to grab the current state of vault and activeCanvas in our nested components.

## Notes
Because of how hyperion works, we do not need to update the context of it using React.Context, but can access it's current state from it. I instantiate a `new Vault();` in our context file in the `defaultState` object and from there we can access this instance `vault`. If calls are made against vault `isLoaded` or `getRef`, the hyperion-framework/vault package acts on these independent of our infrastructure.


src/context/viewer-context.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const [state, dispatch] = React.useReducer(viewerReducer, initialState);
+
+  return (
+    <ViewerStateContext.Provider
+      value={{
+        manifestId,
+        activeCanvas: state.activeCanvas,
+        vault: state.vault,
+        isLoaded: state.isLoaded,
+      }}
```
adamjarling, line 52: @mathewjordan Since we mostly pushing `state.someProperty` into the Context's `value`, can we simplify this and just do:

`<ViewerStateContext.Provider value={state} ..../>` ?     Do we need `manifestId` in the Context values or is that just a holdover from earlier?

src/context/viewer-context.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+function useViewerDispatch() {
+  const context = React.useContext(ViewerDispatchContext);
+  if (context === undefined) {
+    throw new Error("useViewerDispatch must be used within a ViewerProvider");
+  }
+  return context;
+}
+
+export { ViewerProvider, useViewerState, useViewerDispatch };
```
adamjarling, line 77: I like the name change to "Viewer" from "Vault"

src/index.tsx
```diff
@@ -1,38 +1,96 @@
...
-const App: React.FC<Props> = ({ id }) => {
-  const [manifest, setManifest] = useState<ManifestNormalized | undefined>();
-  const [loaded, setLoaded] = useState(false);
+  /**
+   * Retrieve state set by the wrapping <ViewerProvider/> and make
+   * the normalized manifest available from @hyperion-framework/vault.
+   */
+  const state = useViewerState();
+  const { manifestId, vault, isLoaded } = state;
+  const [loaded] = useState(false);
```
adamjarling, line 35: Do we need this?

src/index.tsx
```diff
@@ -1,38 +1,96 @@
...
       .catch((error) => {
         console.error(`Manifest failed to load: ${error}`);
+      })
+      .finally(() => {
+        dispatch({
+          type: "updateIsLoaded",
+          isLoaded: true,
+        });
       });
   }, [loaded]);
```
adamjarling, line 62: I think this could work just as well with an empty [ ] (instead of cueing from `loaded` change in value).  An empty array argument to `useEffect()` means it will only run once after the component has rendered.

src/components/Media/Media.tsx
```diff
@@ -1,35 +1,88 @@
...
+            if (
+              annotation.target === item.id &&
+              contentResourceType.includes(contentResource.type)
+            )
+              return (
+                <MediaItem
+                  key={key}
+                  canvas={canvas}
+                  active={true}
+                  onChange={(canvasId) => {
```
adamjarling, line 67: I think this could just be `onChange={onChange}`?   Just semantics, but sometimes more clear to name our functions unique from React function names.  So maybe our `onChange()` becomes `handleChange()`?

src/components/Media/MediaItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { getLabel } from "../../services/iiif";
+import { CanvasNormalized } from "@hyperion-framework/types";
+
+interface Props {
+  canvas: CanvasNormalized;
+  active: boolean;
+  onChange: (arg0: string) => void;
+}
+
+const MediaItem: React.FC<Props> = ({ canvas, onChange }) => {
```
adamjarling, line 12: Do we want to destructure `active` here as well and grab its value?

adamjarling (approved):
Looks good.  A few minor comments but I think this is a pretty good chunk of code to get into `main` and keeping building on top of.  Nice work.

PR #11, opened 2021-09-29
Render Media Component

mathewjordan:
This pull request lays the ground work for rendering out canvases as individual `<MediaItem/>` in the `<Media/>` component. 

image

It introduces a function `getCanvasByCriteria()` that we may want to treat as a _hook_. This filters out canvases that do not match a specified W3C annotation motivation ("painting", "supplementing"), or content resource type ("Image", "Sound", "Video").

Additionally, some work was completed to indicate to the end user visually which active media item is set.

### DIFF
--- a/src/components/Media/Media.tsx
+++ b/src/components/Media/Media.tsx
@@ -41,9 +41,9 @@ const Media: React.FC<MediaProps> = ({ items, activeItem }) => {
         if (canvasEntity !== undefined)
           return (
             <MediaItem
-              key={key}
-              canvasEntity={canvasEntity}
               active={true}
+              canvasEntity={canvasEntity}
+              key={key}
               handleChange={handleChange}
             />
           );
--- a/src/components/Media/Media.tsx
+++ b/src/components/Media/Media.tsx
@@ -37,8 +37,7 @@ const Media: React.FC<MediaProps> = ({ items, activeItem }) => {
           paintingType,
         );
 
-        let isActive = false;
-        if (activeCanvas === item.id) isActive = true;
+        let isActive = activeCanvas === item.id ? true : false;
 
         if (canvasEntity !== undefined)
           return (
--- a/src/components/Media/Media.tsx
+++ b/src/components/Media/Media.tsx
@@ -28,7 +28,7 @@ const Media: React.FC<MediaProps> = ({ items, activeItem }) => {
 
   const displayItems = items.map((item: object, key: number) => {
     // this probably needs to be written in a .filter()
-    const canvasEntity: object = getCanvasByCriteria(
+    const canvasEntity = getCanvasByCriteria(
       vault,
       item,
       motivation,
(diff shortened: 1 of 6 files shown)


mathewjordan:
Hey @adamjarling I'm not seeing a way to connect this to this issue https://github.com/nulib/repodev_planning_and_docs/issues/2222. Is there something we need to do in the github settings for this repo?

adamjarling:
@mathewjordan Yea, I'm not seeing this either.  @davidschober Do we need to configure this repo somehow so it can see issues from `repodev_planning_and_docs`?

adamjarling:
Ok, I just deleted and re-added `react-media-player` from our Workspace, and now we can see where to connect issues to PRs, and in general appears to be wired up.

mathewjordan:
Thanks @adamjarling . Looks good now.

src/components/Media/Media.tsx
```diff
@@ -35,40 +29,27 @@ const Media: React.FC<MediaProps> = ({ items, activeItem }) => {
   return (
     <MediaWrapper>
       {items.map((item: object, key: number) => {
```
adamjarling, line 31: I'd kind of like to break the code logic out of what's being `return()d`.  It's easier for my brain to read.  What if after `handleChange` declaration, you created a new variable like `const displayItems = items.map(......)`;  and then the return below looks something like:

```
return (
  <MediaWrapper>
    {displayItems.map(item => <MediaItem ... />)}
  </MediaWrapper>
)

Something like that?
```

src/components/Media/Media.tsx
```diff
@@ -35,40 +29,27 @@ const Media: React.FC<MediaProps> = ({ items, activeItem }) => {
...
+        let isActive = false;
+        if (activeCanvas === item.id) isActive = true;
+
+        if (canvasEntity !== undefined)
+          return (
+            <MediaItem
+              active={isActive}
+              canvasEntity={canvasEntity}
+              thumbnail={getThumbnail(vault, canvasEntity, 200, null)}
+              key={key}
```
adamjarling, line 49: This `key` value is coming from the `map` function's `index` value?   If so, and we have a unique id or piece of data for each `item`, lets try and use that instead of using array indexes as keys.

src/components/Media/Media.tsx
```diff
@@ -35,40 +29,27 @@ const Media: React.FC<MediaProps> = ({ items, activeItem }) => {
...
-        const canvas: CanvasNormalized = vault.fromRef(item);
-        // 2. get annotationPage
-        const annotationPage: AnnotationPageNormalized = vault.fromRef(
-          canvas.items[0],
-        );
-        // 3.get annotations
-        const annotations: Array<AnnotationNormalized> = vault.allFromRef(
-          annotationPage.items,
+        // this probably needs to be written in a .filter()
+        const canvasEntity: object = getCanvasByCriteria(
```
adamjarling, line 33: Can we type `canvasEntity` with it's interface from the service helper, instead of using `object`?

adamjarling (changes requested):
All in all looks great.  One idea down in the Media component below...lemme know your thoughts?

davidschober:
So this is all good right?

src/components/Media/Media.tsx
mathewjordan: Good call. I just removed the object type and it now maintains the CanvasEntity type returned from the helper. 
image

adamjarling (approved):
Looks good.

PR #12, opened 2021-09-30
Update render of media

mathewjordan:
- Init helper func to return normalized data.
- Refactor filter of media.
- Return qualified canvases.
- Refine semantics.
- Refine language further.
- Add placeholder for supplementing.
- Update order.
- Initiate getThumbnail method.
- Convertion time from seconds to clean time.
- Style duration.
- Pass active prop to active media item.
- Style active media item.
- Modernize isActive.
- Declutter media return.
- Render key from unique canvas id.
- Maintain type of canvas entity.
- Cleanup logic on media mapping further.
- Removed unneeded params and imports.
- Tidy up media and convert mediaItems from state to const array.

### DIFF
--- a/src/components/Media/Media.tsx
+++ b/src/components/Media/Media.tsx
@@ -41,9 +41,9 @@ const Media: React.FC<MediaProps> = ({ items, activeItem }) => {
         if (canvasEntity !== undefined)
           return (
             <MediaItem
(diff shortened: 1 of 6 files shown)


mathewjordan:
This is junk, closing.

PR #14, opened 2021-10-01
Initial switching of videos in the Viewer on user clicks

adamjarling:
This PR allows for clicking a Media Item thumbnail and the Player will display either a media file or a placeholder for the image (Image doesn't currently pull through).


mathewjordan (approved):
You were right, the `tsc` warnings display but otherwise it looks good to me. :shipit:

PR #15, opened 2021-10-04
Refine typescript definitions and resolve tsc errors

mathewjordan:
This pull request addresses all typescript issues shown on `npm run dev`, `npm run build`, and `tsc`. My resolutions of these issues (40-50 depending on the cascade) largely were resolved by consolidating our types from the @hyperion-framework. We still may want to be a little more specific in the future with our definitions, but that work can be left to more focused issues.


mathewjordan:
@adamjarling This is ready for your review now.

PR #19, opened 2021-10-06
Update some defensive code around parsing the manifest

adamjarling:
Round one of some code which will help keep the app from crashing if it runs into parsing errors.

### To test
I committed a file in /public which is a bad manifest.  You can feed that into `dev.tsx` as the `manifestId` to render the bad manifest and see if the app loads.

### DIFF
--- a/src/components/Media/MediaItem.tsx
+++ b/src/components/Media/MediaItem.tsx
@@ -8,14 +8,14 @@ import {
   InternationalString,
 } from "@hyperion-framework/types";
 
-interface Props {
+export interface MediaItemProps {
   canvas: CanvasNormalized;
   active: boolean;
-  thumbnail: IIIFExternalWebResource;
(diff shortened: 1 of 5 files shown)


mathewjordan (approved):
This works for me, renders out a fairly messy layout but definitely better than a fatal. :shipit: 

image

src/components/Media/MediaItem.tsx
```diff
@@ -31,7 +31,7 @@ const MediaItem: React.FC<Props> = ({
     >
       <figure>
         <div>
-          <img src={thumbnail.id} />
+          {thumbnail?.id && <img src={thumbnail.id} />}
```
mathewjordan, line 34: Love this.

adamjarling: Yea, one-liner conditionals make life a little easier eh?

PR #20, opened 2021-10-06
2224 render parsed vtt

mathewjordan:


src/components/Navigator/Navigator.tsx
```diff
@@ -1,73 +1,54 @@
...
     <NavigatorWrapper data-testid="navigator-wrapper">
       <NavigatorHeader>
-        <NavigatorTab label="Camptium" active={true} />
-        <NavigatorTab label="Mauris" active={false} />
-        <NavigatorTab label="Tristique" active={false} />
+        {resources &&
+          resources.map((resource) => (
+            <NavigatorTab
+              resource={resource}
+              active={activeResource === resource.id ? true : false}
```
adamjarling, line 36: Could tighten this up to `active={activeResource === resource.id}`

src/components/Navigator/NavigatorResource.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useEffect } from "react";
+import { fromVtt } from "subtitles-parser-vtt";
```
adamjarling, line 2: Does this package have any @types we can import via [Definitely Typed](https://www.typescriptlang.org/dt/search?search=)?

adamjarling: And after checking out this package `https://www.npmjs.com/package/subtitles-parser-vtt`, curious if there are any more mature options to pull data out of our WebVTTs?   I'm fine either way, but a package with roughly 100 weekly downloads with a version of `0.1.0` could conceivably go away, not get updated, more bug-prone, etc.   Your call, just thinking out loud.

src/components/Navigator/NavigatorTab.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+interface NavigatorTabProps {
+  active: boolean;
+  resource: LabeledResource;
+  handleChange: (arg0: string) => void;
+}
+
+const NavigatorTab: React.FC<NavigatorTabProps> = ({
+  active,
+  resource,
```
adamjarling, line 15: How heavy to you imagine `resource` to be?   If a WebVTT contains a transcript of thousands of lines say, would that content come along with the value for `resource`?   Just wondering if we could only pass in `id` and `label` off `resource`...

src/hooks/use-hyperion-framework/getContentResourcesByCriteria.ts
```diff
@@ -0,0 +1,44 @@
+import {
+  AnnotationPage,
+  CanvasNormalized,
+  InternationalString,
+} from "@hyperion-framework/types";
+
+export interface LabeledResource {
```
adamjarling, line 7: Maybe we make a breadcrumb comment here for later to build on top of Hyperion's "ContentResource" type, or whatever it was we were discussing earlier, and figure out concatenating existing types with our own?

adamjarling (approved):
Looks awesome all around.  Nice work.  A few comments for thought...

PR #21, opened 2021-10-06
Commit the /dist build folder to repo so we can import directly from …

adamjarling:
…Github

@mathewjordan This PR commits our build `/dist` folder so we can import the app into any consuming app before it's ready for NPM.


mathewjordan (approved):
Sounds good to me.

PR #23, opened 2021-10-11
Refine styling of RMP.

mathewjordan:
## What does this do?

This pull request refines the styling of the RMP app:
- adds variable based colors, fonts, and transitions using stitches theming, naming colors with the thought that these might someday be used for multiple _modes_.
- defines font sizes for various selectors resolving small font sizes issues for the media item labels and the navigator cue text.
- adds a pure CSS spinning indicator for active cue item (this is just styled for now and set to the first child) actual work for active cue item will occur in [2188](https://github.com/nulib/repodev_planning_and_docs/issues/2188)
- adds a pure CSS play indicator on hover of a navigator cue
- includes the Northwestern font imports in the `public/index.html` for a better dev experience

##  Notes
If you have thoughts on a better location for `theme.tsx`, I'm all ears. Long term this could ultimately house a light and a dark mode though.

image

### DIFF
--- a/src/components/Media/MediaItem.tsx
+++ b/src/components/Media/MediaItem.tsx
@@ -7,6 +7,7 @@ import {
   IIIFExternalWebResource,
   InternationalString,
 } from "@hyperion-framework/types";
+import { theme } from "theme";
 
 export interface MediaItemProps {
   canvas: CanvasNormalized;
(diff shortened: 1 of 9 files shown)


src/components/Media/MediaItem.tsx
```diff
@@ -7,6 +7,7 @@ import {
   IIIFExternalWebResource,
   InternationalString,
 } from "@hyperion-framework/types";
+import { theme } from "theme";
```
adamjarling, line 10: I really like this

PR #37, opened 2021-11-15
Add initial config option to RMP outer component.  Fix up jest tests …

adamjarling:
…to pass

Allow an outer config `options` prop to be sent into the RMP. 

`<App manifestId={sampleManifest} options={{ showTitle: false }} />`

 The only currently supported option is `showTitle`, which will default to `true` unless set to `false` in config.  

image


src/components/Navigator/Navigator.tsx
```diff
@@ -39,9 +39,9 @@ export const Navigator: React.FC<NavigatorProps> = ({
       value={activeResource}
     >
       <List aria-label="select chapter" data-testid="navigator-list">
-        {resources.map((resource) => (
-          <Trigger value={resource.id as string}>
-            {getLabel(resource.label as InternationalString, "en")}
+        {resources.map(({ id, label }) => (
+          <Trigger key={id} value={id as string}>
```
mathewjordan, line 43: Looks like we both put in a fix here. I'll let you decide how to handle. Yours is de-structured so I lean that way.

mathewjordan (approved):
:shipit: Looks great. Thanks for adding those tests on `<Player />` too.

src/components/Navigator/Navigator.tsx
adamjarling: Cool.  Let's keep the destructured approach.  Easier to read maybe?

PR #41, opened 2021-11-24
2418 audio visualizer

adamjarling:
This PR introduces an audio visualizer component, which displays moving visual bars which are driven by audio stream data.

### DIFF
--- a/src/components/Player/AudioVisualizer.styled.tsx
+++ b/src/components/Player/AudioVisualizer.styled.tsx
@@ -1,0 +1,5 @@
+import { styled } from "stitches";
+
+export const AudioVisualizerWrapper = styled("canvas", {
+  position: "absolute",
+});
--- a/src/components/Player/Player.styled.tsx
+++ b/src/components/Player/Player.styled.tsx
@@ -13,5 +13,16 @@ export const PlayerWrapper = styled("div", {
     width: "100%",
     height: "100%",
     backgroundColor: "$primaryAlt",
+    zIndex: "1",
+  },
+
+  variants: {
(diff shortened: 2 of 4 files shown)


src/components/Player/Player.styled.tsx
```diff
@@ -1,17 +1,19 @@
 import { styled } from "stitches";
 
 export const PlayerWrapper = styled("div", {
+  backgroundColor: "$primaryAlt",
```
mathewjordan, line 4:
```suggestion
  position: "relative",
  backgroundColor: "$primaryAlt",
```

src/components/Player/AudioVisualizer.styled.tsx
```diff
@@ -0,0 +1,5 @@
+import { styled } from "stitches";
+
+export const AudioVisualizerWrapper = styled("canvas", {
+  position: "absolute",
```
mathewjordan, line 4:
```suggestion
  position: "absolute",
  width: "100%",
  height: "100%",
```

mathewjordan (changes requested):
Looks great @adamjarling except that the canvas does not expand or contract to fit the container if the width/height of the player wrapper changes. I think all we need is a `position: relative;` following by some explicit width/height attributes.

image

image

image

src/components/Player/Player.styled.tsx
adamjarling: Good call.

mathewjordan (approved):
May be a few more touch ups for styling the bars but I like it.

PR #42, opened 2021-12-03
Add resource type icons to thumbnails.

mathewjordan:
### Description
This pull request adds work/resource type icons to the media thumbnails for Audio (Sound), Image, and Video. To implement, I merged it with the previously existing `<Duration/>` component. Now all canvas types will have some sort of `<Type>`  with an embedded `<Icon>` (from `@nulib/design-system`), and Audio/Video works will show duration as a label.

image

- Add resource type icons to Media thumbnails.
- Add in some transparency.
- Add IconPath and switch statement.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,17 +13,18 @@
         "@hyperion-framework/types": "^1.0.0",
         "@hyperion-framework/validator": "^1.0.0",
         "@hyperion-framework/vault": "^1.0.2",
+        "@nulib/design-system": "^1.3.5",
         "@radix-ui/react-radio-group": "^0.1.1",
         "@radix-ui/react-tabs": "^0.1.1",
         "@stitches/react": "^1.2.0",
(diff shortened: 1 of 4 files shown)


adamjarling (approved):
Looks great.  One thing that I was wondering about (either in this or the design system repo), is whether we're going to run into weird JS compilation/build bugs in the future using `type` or `Type` as variable names?   Like violating a reserved word policy some libraries/packages/IDEs might have?

mathewjordan:
> Looks great. One thing that I was wondering about (either in this or the design system repo), is whether we're going to run into weird JS compilation/build bugs in the future using `type` or `Type` as variable names? Like violating a reserved word policy some libraries/packages/IDEs might have?

You're right, we may want to clean that up. In this case, both could be cleaned up as `resourceType` and `ResourceType` as it refers to the IIIF resource. Maybe we should have a cleanup issue to go through and prune all this up?

PR #45, opened 2021-12-06
Baseline responsiveness.

mathewjordan:
### Description

The work included here adds baseline responsive behavior to the project. Most of this is based around a specific breakpoint of `sm` defined in the stitches configuration file. When this breakpoint is reached and true, the navigator (if present) collapse under the Player and Media elements. 

By default the navigator will not be displayed, however, toggle with the title **View Navigator** will be shown to the user on canvases where the Navigator is available. When expanded, the player goes into a emulated modal state where navigator contents are shown and scrollable. This behavior exists to limit the possibility of multiple scrollbars. When the user clicks **Media Items**, scrolling functionality on the consuming page is present once again.

image

image

### Notes
Included as well are various touchups to refine font-sizing project wide and thumbnail sizing in the media tray at the `sm` viewport. This project now directly imports the Radix UI element **Collapsible**. Our use of it for the navigator toggle functionality is rather boutique and I don't image that we'd want to leverage a potential design-system version of this.

### DIFF
--- a/src/components/Viewer/Viewer.tsx
+++ b/src/components/Viewer/Viewer.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import {
   ExternalResourceTypes,
   IIIFExternalWebResource,
@@ -19,9 +19,18 @@ import {
   ViewerWrapper,
   ViewerInner,
   Main,
+  CollapsibleTrigger,
+  CollapsibleContent,
+  MediaWrapper,
   Header,
   Aside,
 } from "./Viewer.styled";
+import { Button } from "@nulib/design-system";
+import { useMediaQuery } from "hooks/useMediaQuery";
+import { useBodyLocked } from "hooks/useBodyLocked";
+import { media } from "stitches";
+import { LabeledResource } from "hooks/use-hyperion-framework/getSupplementingResources";
+import * as Collapsible from "@radix-ui/react-collapsible";
 
 interface ViewerProps {
   manifest: ManifestNormalized;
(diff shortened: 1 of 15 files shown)


src/components/Viewer/Viewer.tsx
```diff
@@ -19,30 +19,64 @@ import {
...
+    }
+    setIsNavigatorOpen(false);
+  }, [isSmallViewport]);
+
+  useEffect(() => {
+    if (!isSmallViewport) {
+      setIsBodyLocked(false);
+      return;
+    }
+    isNavigatorOpen ? setIsBodyLocked(true) : setIsBodyLocked(false);
```
adamjarling, line 70: No biggie, but could write this `setIsBodyLocked(isNavigatorOpen)`

src/components/Viewer/Viewer.tsx
```diff
@@ -51,45 +85,62 @@ const Viewer: React.FC<ViewerProps> = ({ manifest, theme }) => {
       );
       setPainting({ ...painting });
     }
+    setResources(resources);
+    resources.length === 0 ? setIsNavigator(false) : setIsNavigator(true);
```
adamjarling, line 89: Same here.... `setIsNavigator(resources.length === 0)`

src/hooks/useBodyLocked.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useLayoutEffect, useState } from "react";
```
adamjarling, line 1: Awesome hook composition!

src/hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from "react";
+
+export const useMediaQuery = (mediaQuery: string) => {
+  const match = () => {
+    if (!window.matchMedia) {
```
adamjarling, line 5: Great use of the browser native API.  Never knew about `matchMedia`...cool.

adamjarling (approved):
This is really good work.  A few small syntax ideas, but no biggie either way.

src/hooks/useBodyLocked.ts
mathewjordan: `useLayoutEffect` was a new one for me too.

src/components/Viewer/Viewer.tsx
adamjarling: Yea, anytime you see 2 identical variables or expressions, there's generally a trick to tighten 'em up.  Less code, easier to read in a few months.

PR #50, opened 2021-12-21
Add IIIF Badge.

mathewjordan:
### What does this do?
This adds an `<svg>` IIIF Badge in the upper right corner. Clicking it will display a `<Popover>` imported from the `@nulib/design-system`. The popover contains two options at the moment. One for linking directly to the manifest in a new tab and another for copying the manifest URL. All functionality should be keyboard accessible using tabbing. The copy functionality includes an added hook `useCopyToClipboard`  and the new component `<CopyText>`.

The IIIF badge can also be disabled using the `options` prop. 
```
<ReactMediaPlayer manifestId={manifestId} options={{ showIIIFBadge: false }} />
```

image

### Other Updates
-  The Design System package had been updated
-  The `<Viewer>` component was broken out into sub-components for a **ViewerHeader**  and **ViewerContent**


src/components/Viewer/IIIFBadge.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+
+const IIIFBadge: React.FC = () => {
+  const red = "#ed1d33";
+  const blue = "#2873ab";
+
+  return (
+    <svg viewBox="0 0 493.35999 441.33334" id="iiif-logo" version="1.1">
```
adamjarling, line 8: One of these days you're going to have to show me how you make these.

src/hooks/useCopyToClipboard.ts
```diff
@@ -0,0 +1,28 @@
...
+      return;
+    }
+
+    const timeoutId = setTimeout(() => setCopyStatus(undefined), notifyTimeout);
+
+    return () => clearTimeout(timeoutId);
+  }, [copyStatus]);
+
+  return [copyStatus, copy];
+};
```
adamjarling, line 28: Nice doing this w/o an extra NPM package.

adamjarling (approved):
This is great.

PR #51, opened 2022-01-06
Request tilesource from iiif service.

mathewjordan:
image

## What does this do?
This pull request resolves and issue where some IIIF Image Services do not redirect to the info.json and thus do not hand off the tilesource json if the image service ID is directly handed to OpenSeadragon. This PR includes code that will request the rewrite the image service ID URI using regex (appending info.json) and then request the tilesource json first, then hand that promised json to OSD. If this fails to resolve, then a console error is displayed. This is similar logic to how Mirador seems to handle this issue.

## How do we review?
This PR includes a sample National Library of Scotland manifest that was previously failing to load in OpenSeadragon. It should now display as expected.

### DIFF
--- a/src/components/ImageViewer/ImageViewer.tsx
+++ b/src/components/ImageViewer/ImageViewer.tsx
@@ -6,6 +6,7 @@ import {
 } from "@hyperion-framework/types";
 import { Navigator, Viewport, Wrapper } from "./ImageViewer.styled";
 import Controls from "./Controls";
+import { getInfoResponse } from "services/iiif";
 
 const ImageViewer: React.FC<IIIFExternalWebResource> = ({ service }) => {
   const [openSeadragonInstance, setOpenSeadragonInstance] = useState<Viewer>();
@@ -33,7 +34,11 @@ const ImageViewer: React.FC<IIIFExternalWebResource> = ({ service }) => {
    * Loads tileSource of current canvas from IIIF image service
    */
   useEffect(() => {
-    if (imageService) openSeadragonInstance?.open(imageService.id);
+    if (imageService) {
+      getInfoResponse(imageService.id).then((tileSource) =>
+        openSeadragonInstance?.open(tileSource),
+      );
+    }
   }, [openSeadragonInstance, imageService]);
 
   /**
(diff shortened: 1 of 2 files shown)


src/components/ImageViewer/ImageViewer.tsx
```diff
@@ -33,7 +34,11 @@ const ImageViewer: React.FC<IIIFExternalWebResource> = ({ service }) => {
    * Loads tileSource of current canvas from IIIF image service
    */
   useEffect(() => {
-    if (imageService) openSeadragonInstance?.open(imageService.id);
+    if (imageService) {
+      getInfoResponse(imageService.id).then((tileSource) =>
+        openSeadragonInstance?.open(tileSource),
```
adamjarling, line 39: Will `tileSource` always have a value here?  What would happen if the shape of `tileSource` is not what we expect, or `undefined` or does OpenSeadragon handle that for us?

src/hooks/getInfoResponse.ts
```diff
@@ -0,0 +1,12 @@
+export const getInfoResponse = (id: string) =>
+  fetch(`${id.replace(/\/$/, "")}/info.json`)
+    .then((response) => response.json())
+    .then((json) => json)
+    .catch((error: any) => {
+      console.error(
```
adamjarling, line 6: Just thinking if this is a hook, and it enters the `catch` block if we should provide a reasonable error response to calling code.  Like `null` or `{}` or something maybe?

adamjarling (changes requested):
Looks good and gets the job done.  A quick question on the return of `getInfoResponse` noted above/below.

Also, if we want to consider this a hook, let's run with hook convention and rename the file and function to something like `useGetInfoResponse`.   This also kinda feels like an API util, so could see it going as-is in `/services/utils.ts`.  Your call.

mathewjordan:
> Looks good and gets the job done. A quick question on the return of `getInfoResponse` noted above/below.
> 
> Also, if we want to consider this a hook, let's run with hook convention and rename the file and function to something like `useGetInfoResponse`. This also kinda feels like an API util, so could see it going as-is in `/services/utils.ts`. Your call.

I moved this to services/iiif.ts. Utils.ts is more for general stuff. Thoughts?

adamjarling (approved):

PR #53, opened 2022-01-13
Add ability to search/filter canvases by label.

mathewjordan:
## What does this do?
This introduces basic search/filter functionality to the Viewer, allowing as user to filter the string value of the canvas label. The search button toggles and auto-focuses an input for a user to input this value. Upon filtering, the media items in the `<Media />` component are filtered and mapped. A close button is rendered allowing the user to toggle this off. Upon closing, the filter value is reset. Hitting the `esc` "Escape" keyboard button will close filter input as well. 

## What this doesn't do.
This filtering is pretty basic and somewhat requires the manifest canvases to have label values useful the the user. This does not filter by index or canvas number, in cases where a user might expect typing "10" would take them to the 10th canvas, unless of course the label for that canvas contained "10" within it's value.

## How do we test this?
A manifest with ~30 canvases and long canvas labels is ready for testing on this branch.
1. Click **Search icon**
2. Type **asmara**
3. Items will filter
4. Try your own

## Future considerations
1. This is not debounced. For performance reasons in the future we may want to debounce at some time interval.
2. The SVG icons in the `<Filter />` component might want to be moved to a central location and handled in some uniform pattern with others in the viewer.

### DIFF
--- a/src/components/Media/Filter.tsx
+++ b/src/components/Media/Filter.tsx
@@ -1,0 +1,66 @@
+import React, { useState } from "react";
+import { Form, Input, Toggle, Wrapper } from "./Filter.styled";
+import useKeyPress from "hooks/useKeyPress";
+
+interface Props {
+  handleFilter: (arg0: String) => void;
+}
+
+const CloseIcon: React.FC = () => {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">
+      <title>Close</title>
+      <path d="M289.94 256l95-95A24 24 0 00351 127l-95 95-95-95a24 24 0 00-34 34l95 95-95 95a24 24 0 1034 34l95-95 95 95a24 24 0 0034-34z" />
+    </svg>
+  );
+};
+
+const SearchIcon: React.FC = () => {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512">
+      <title>Search</title>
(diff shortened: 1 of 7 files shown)


src/components/Media/Filter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) =>
+    handleFilter(event.target.value);
+
+  return (
+    <Wrapper isToggle={toggleFilter}>
+      <Form>
+        {toggleFilter && (
+          <Input
+            autoFocus
+            onChange={(event) => handleChange(event)}
```
adamjarling, line 58: You can just pass the function by reference here:   

`onChange={handleChange}`

src/components/Media/Filter.tsx
```diff
@@ -0,0 +1,70 @@
...
+            placeholder="Search"
+          />
+        )}
+        <Toggle onClick={() => handleToggle()} type="button">
```
adamjarling, line 62: And here:

`onClick={handleToggle}`

src/hooks/useKeyPress.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @param {string} key - the name of the key to respond to, compared against event.key
+ * @param {function} action - the action to perform on key press
+ */
+const useKeyPress = (key, action) => {
+  useEffect(() => {
+    function onKeyup(e) {
+      if (e.key === key) action();
+    }
+    window.addEventListener("keyup", onKeyup);
+    return () => window.removeEventListener("keyup", onKeyup);
```
adamjarling, line 13: Nice cleanup.  Once we know this hook works properly, could put this in the common hooks repo too.

adamjarling (approved):
Looks great.  Small comment below, but everything works and looks great.

PR #54, opened 2022-01-14
Support VTT nested chapters, and updates handling current time as a React context

adamjarling:
### Description
This PR:

- Set up webVtt hook and handle initial parsing of flat structure into nested cues structure
- Update to complete nested structure from a flat array.  Extends beyond one child.
- First pass at nested UI display
- Update styling for nested UI display
- Logging code to track down why the Radix Tab Content component is blocking a new Resource from displaying, when changing active canvas.
- Fix potential orphan childed if parent end times are equal.
- Remove background styling on nested children.
- Remove console logs.

image

It also pulls out the state variable keeping track of the Video player's `currentTime`, and puts it into its own Context.  This reduces top-level component renders, and moves the continuous rendering out to primarily the `Cue` component.  Somewhat of a performance optimization to be realized if WebVTT files get huge.

### To Test
In the current PR Code, click on the 4th Video in the Tray ("Video canvas, non-streaming (.mp4), with supplementing annotations as WebVTT"), to see nested navigation in the UI. 

You'll also notice in the React Dev Tools Profiler tab, a reduced rendering time, who's Flamegraph used to look like:

image

and now looks much more like:

image

Pushing all the rendering down to really quick renders.


src/hooks/use-webvtt.ts
```diff
@@ -0,0 +1,115 @@
+import { v4 as uuidv4 } from "uuid";
```
mathewjordan, line 1: This is screaming at us when doing `tsc`, not sure how to best handle.

mathewjordan (changes requested):
Awesome, thanks for cleaning up the currentTime tracking.

Looks like we can now removed `useState` at https://github.com/nulib/react-media-player/blob/8b109cac666cc384c4634406cd4fc7308e199a61/src/components/Viewer/Content.tsx#L1

src/hooks/use-webvtt.ts
adamjarling: Gotcha.  I'll import the `@types/uuid` package as a dev dependency.

adamjarling:
@mathewjordan Just pushed up requested changes

mathewjordan (approved):
👍

PR #55, opened 2022-01-18
Add functionality for rendering VTTs with nested cues

mathewjordan:
- Set up webVtt hook and handle initial parsing of flat structure into nested cues structure
- Update to complete nested structure from a flat array.  Extends beyond one child.
- First pass at nested UI display
- Update styling for nested UI display
- Logging code to track down why the Radix Tab Content component is blocking a new Resource from displaying, when changing active canvas.
- Fix potential orphan childed if parent end times are equal.
- Remove background styling on nested children.
- Remove console logs.


adamjarling:
Cherry picked @mathewjordan commits to https://github.com/nulib/react-media-player/pull/54

PR #56, opened 2022-01-18
Wrap the Viewer in an Error Boundary

adamjarling:
This PR wraps the main Viewer in a React Error Boundary to prevent non-network related crashes of the app in the browser.   If there's a need in the future, we can wrap parts of the app in their own Error Boundaries (ie. Player, Media Tray, Navigator, etc.)

image

### To Test:
In some child component, initiate a syntax error, or insert a `throw new Error("hey I just errored")`.  Notice the Error Boundary catches the error.


mathewjordan (approved):
Love this.

PR #57, opened 2022-01-20
Add Next/Previous functionality.

mathewjordan:
## What does this do?
This pull request adds next and previous functionality to the viewer.

- The Filter component has been renamed Controls as it now contains both directional navigation for next/previous as well as the search/filtering functionality.
- The Media tray is now disabled if the length of canvases is `1`.
- ScrollTo functionality on the active thumbnail has been included for a better user experience.
- Some minor styling updates (slightly adjusted blur) have been made to thumbnails to better offset them from adjacent elements in the Controls component.
- Next/Previous buttons are hidden when search filtering is toggled on.
- Line height of canvas labels is now capped at `5` with an ellipsis. 
- SVG Icons are still housed within the Control component and should probably be moved to a more central location at some point in the future


adamjarling (approved):
This is awesome!!!  Fantastic work.  Super slick.

PR #58, opened 2022-01-31
Disable autoplay on initial canvas load.

mathewjordan:
## Summary

This will disable autoplay for Sound and Video canvases on initial render. It does so by not running `play()` if **startTime** is _0_. It retains the ability to automatically play if a user clicks a cue in the navigator.


adamjarling (approved):
Looking good.

PR #59, opened 2022-02-01
Add captioning and basic filtering configuration.

mathewjordan:
## Summary
This pull request refines captioning behavior for HTML 5 video tracks and allows an implementing developer to selectively ignore supplementing canvases that may not be intended for closed captioning/subtitling. A new configuration option of `ignoreCaptionLabels` provides an array of string values that will filter and ignore rendering of the `<track/>`. To make things even cleaner, the `<track/>` element is now a `<Track/>` sub component.

## Testing
Test by filtering tracks to the matching label of the annotation resource.

### Example:
`
const options: any = {
  ignoreCaptionLabels: ["Chapters"],
};
`

Where the annotation intended to ignore has the structure:
```
{
  "id": "http://127.0.0.1:8080/fixtures/iiif/manifests/captions.json/canvas/1/page/annotation_page/1/annotation/2",
  "type": "Annotation",
  "motivation": "supplementing",
  "body": {
    "id": "http://127.0.0.1:8080/fixtures/vtt/around_the_corner_chapters.vtt",
    "type": "Text",
    "format": "text/vtt",
    "label": {
      "en": ["Chapters"]
    },
    "language": "en"
  },
  "target": "http://127.0.0.1:8080/fixtures/iiif/manifests/captions.json/canvas/1"
}
```

## Note
While completing local testing, I did not that captioning was not working in Mozilla Firefox Developer edition. It does seem to work in all major browsers including the consumer level firefox.


src/components/Player/Track.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { InternationalString } from "@hyperion-framework/types";
+import { LabeledResource } from "hooks/use-hyperion-framework/getSupplementingResources";
+import { getLabel } from "hooks/use-hyperion-framework";
+
+export interface TrackProps {
+  resource: LabeledResource;
+  ignoreCaptionLabels: string[];
```
adamjarling, line 8: Could we make `ignoreCaptionLabels` optional?  And then when we destructure the props, set a default value below like:

```
const Track: React.FC<TrackProps> = ({ resource, ignoreCaptionLabels = [] }) => {
```

src/context/viewer-context.tsx
```diff
@@ -3,11 +3,14 @@ import { Vault } from "@hyperion-framework/vault";
...
 export type ConfigOptions = {
   showTitle: boolean;
+  showIIIFBadge: boolean;
+  ignoreCaptionLabels: string[];
 };
 
 const defaultConfigOptions = {
   showTitle: true,
   showIIIFBadge: true,
+  ignoreCaptionLabels: [],
```
adamjarling, line 13: Ah, well if you're setting a default value here, then my above comment might not matter.

adamjarling (approved):
Looks good.

src/components/Player/Track.tsx
mathewjordan: Yeah, I did this so that the type def was always an array.

PR #60, opened 2022-02-04
2416 audio placeholder canvas

adamjarling:
@mathewjordan Would like to chat on this...about 1.) Hyperion shortcut parsing and 2.) Setting up hooks and leaf level components to be tested easier (ie. maybe not passing in `vault`, which is hard if not impossible to mock in tests).

This PR supports Audio placeholder images when pulled in via `accompanyingCanvas` in a IIIF Manifest

### DIFF
--- a/public/fixtures/iiif/manifests/audio-accompanying-canvas.json
+++ b/public/fixtures/iiif/manifests/audio-accompanying-canvas.json
@@ -1,0 +1,75 @@
+{
+  "@context": "http://iiif.io/api/presentation/3/context.json",
+  "id": "https://iiif.io/api/cookbook/recipe/0014-accompanyingcanvas/manifest.json",
+  "type": "Manifest",
+  "label": {
+    "en": ["Partial audio recording of Gustav Mahler's _Symphony No. 3_"]
+  },
+  "items": [
+    {
+      "id": "https://iiif.io/api/cookbook/recipe/0014-accompanyingcanvas/canvas/p1",
+      "type": "Canvas",
+      "label": {
+        "en": ["Gustav Mahler, Symphony No. 3, CD 1"]
+      },
+      "duration": 1985.024,
+      "accompanyingCanvas": {
+        "id": "https://iiif.io/api/cookbook/recipe/0014-accompanyingcanvas/canvas/accompanying",
+        "type": "Canvas",
+        "label": {
+          "en": ["First page of score for Gustav Mahler, Symphony No. 3"]
+        },
+        "height": 998,
+        "width": 772,
+        "items": [
+          {
+            "id": "https://iiif.io/api/cookbook/recipe/0014-accompanyingcanvas/canvas/accompanying/annotation/page",
+            "type": "AnnotationPage",
+            "items": [
+              {
+                "id": "https://iiif.io/api/cookbook/recipe/0014-accompanyingcanvas/canvas/accompanying/annotation/image",
+                "type": "Annotation",
+                "motivation": "painting",
+                "body": {
+                  "id": "https://iiif.io/api/image/3.0/example/reference/4b45bba3ea612ee46f5371ce84dbcd89-mahler-0/full/,998/0/default.jpg",
+                  "type": "Image",
+                  "format": "image/jpeg",
+                  "height": 998,
+                  "width": 772,
+                  "service": [
+                    {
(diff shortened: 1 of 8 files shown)


mathewjordan:
@adamjarling Sorry, I missed this. Did you still want to chat through this at some point?

adamjarling:
@mathewjordan Yea, I would like to go over it for 10 minutes.  Whenever you want, lemme know via Slack.

mathewjordan:
This might be the order of how we do things for all canvases including accompanyingCanvas.

### if thumbnail
- Does it have a `thumbnail`? Yes, then use it
  - Does thumbnail have a `service`? Yes, use it.
  - If no service, use `id`

### if no thumbnail
- Use items array and get motivation painting
  - Does thumbnail have a `service`? Yes, use it.
  - If no service, use `id`

adamjarling:
Got it.  @mathewjordan Pushed up some changes which enhances the parsing a bit.  Still running into a few Type issues:
- Should `accompanyingCanvas` be of Hyperion type `Canvas`?  and
- Should the value of `thumbnail` or `body` be of type `IIIFExternalWebResource`?

For example, TypeScript was yelling at me when it encountered `id` on a `service[0]` object, however looked like it would have accepted `@id`.  

I got around it by typing as `:any` for now and put a TODO: comment next to each as a reminder to address this in the larger clean up all tests Github issue, where I can maybe mentally get into it for a few hours and understand the type mappings better.

mathewjordan:
@adamjarling I'm wondering about the poster attribute approach and how this might conflict with a visualizer/waveform. I have some thoughts on how we can mimic the poster feel but then tuck it away while the audio is playing.

public/fixtures/iiif/manifests/audio-accompanying-canvas.json
```diff
@@ -0,0 +1,75 @@
...
+          "items": [
+            {
+              "id": "https://iiif.io/api/cookbook/recipe/0014-accompanyingcanvas/canvas/page/annotation/segment1-audio",
+              "type": "Annotation",
+              "motivation": "painting",
+              "body": {
+                "id": "https://fixtures.iiif.io/audio/indiana/mahler-symphony-3/CD1/medium/128Kbps.mp4",
+                "type": "Sound",
+                "duration": 1985.024,
+                "format": "video/mp4"
```
mathewjordan, line 66: This is a video file but of type sound, is that expected?

src/components/Player/Player.tsx
```diff
@@ -16,14 +20,28 @@ interface PlayerProps {
 }
 
 const Player: React.FC<PlayerProps> = ({ painting, resources }) => {
-  const viewerState: any = useViewerState();
-  const { configOptions } = viewerState;
-
   const playerRef = React.useRef(null);
   const isAudio = painting?.format?.includes("audio/");
```
mathewjordan, line 24: Not your code here but I'm wondering if this should be...
```suggestion
  const isAudio = painting?.type === "Sound";
```

adamjarling:
@mathewjordan Yea, sounds good to me.  Any ideas you have, feel free to lemme know or run with it if you like.

src/components/Player/Player.tsx
```diff
@@ -109,6 +127,7 @@ const Player: React.FC<PlayerProps> = ({ painting, resources }) => {
         width={painting.width}
         onPlay={handlePlay}
         crossOrigin="anonymous"
+        poster={posterImage}
```
mathewjordan, line 130: I really like the simplicity of it but have concerns that this would conflict with visualizer/waveform.

public/fixtures/iiif/manifests/audio-accompanying-canvas.json
adamjarling: Ah, good catch.  I copied and pasted directly from the IU recipe, but seems a `mp4` file can legally be audio or video, so guess this is ok?   Will our code react unexpectedly against this?

src/components/Player/Player.tsx
adamjarling: Yea, sounds good to me.  Guessing the original line is looking at some form of `mime type`, but an authored `painting.type` seems like it should be trusted more eh?

mathewjordan (approved):
Sorry this has sat so long. Let's move get this work in and I'll create a new issue for handling the media in the `poster` attribute in a way that doesn't impact waveform/visualization.

PR #64, opened 2022-03-14
2710 create static site hosting

adamjarling:
@mathewjordan I'll need to get this merged in before being able to test if it actually works or not.  So this PR comes in 2 parts:

1. You can now generate a static version of the site by running `npm run build:static`.  This will create a static version of the dev environment, so whatever IIIF manifest is being sent into the component, is what the static site will generate as well.
2. Start up a local server (ie. `npx serve`) and navigate to the `/static` directory to verify if serves ok for you.

### Static build notes
I had to manually wire this up, and what happens is it copies over all the files, except ESBuild's `script.js` bundled file, from `/public` to `/static`.  This currently is `/fixtures` and `index.html`.

Then ESBuild bundles the code, using a few environmental variables, and puts the `script.js` into the `/static` folder.

image

~~Next steps will be on merge to see if it triggers a build to the `gh-pages` branch.~~

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,2 @@
+DEV_URL=http://127.0.0.1:8080
+STATIC_URL=.
--- a/.github/workflows/gh-pages.deploy.yml
+++ b/.github/workflows/gh-pages.deploy.yml
@@ -1,0 +1,35 @@
+name: Deploy to GitHub Pages
+
+on:
+  push:
+    branches: [main]
+
+jobs:
+  build:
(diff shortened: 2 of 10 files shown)


adamjarling:
@mathewjordan I pushed up a commit which pointed Github Actions to the current branch, and everything built successfully.  So when I merge this, I'll point the script back to `main` and we should be good to go.

https://samvera-labs.github.io/clover-iiif/

PR #65, opened 2022-03-16
Update some of the test failures, but not complete.

adamjarling:
Partial PR to address low hanging fruit test fixes.  The others may require a bit more though on the best approach.


mathewjordan (approved):
Looks good. Seeing just two failing right now.

PR #72, opened 2022-04-07
Update Vault and IIIF type defs.

mathewjordan:
## What does this do?

This PR gets Clover off of the old `@hyperion-framework` and use the newer `@iiif/vault` and `@iiif/presentation-3` packages instead. Some logic required updating throughout due to soome small changes during the package extraction/migration to IIIF Commons. See https://github.com/IIIF-Commons/vault/#changes-to-hyperion-vault for that.

Otherwise, things should be functioning pretty much the same. All manifests that previously rendered seem to be loading and rendering as expected. Let me know if you see otherwise.


src/components/Player/Player.test.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
 import { render } from "@testing-library/react";
 import Player from "./Player";
-import { IIIFExternalWebResource } from "@hyperion-framework/types";
-import { LabeledResource } from "../../hooks/use-hyperion-framework/getSupplementingResources";
+import { IIIFExternalWebResource } from "@iiif/presentation-3";
+import { LabeledResource } from "../../hooks/use-iiif/getSupplementingResources";
```
adamjarling, line 5: Mind swapping out the relative path here, and anywhere we come across them?

src/components/Media/Media.tsx
```diff
@@ -1,19 +1,15 @@
 import React, { useEffect, useState } from "react";
 import { Group } from "./Media.styled";
-import { CanvasEntity } from "hooks/use-hyperion-framework/getCanvasByCriteria";
-import {
-  getCanvasByCriteria,
-  getLabel,
-  getThumbnail,
-} from "hooks/use-hyperion-framework";
+import { CanvasEntity } from "hooks/use-iiif/getCanvasByCriteria";
```
adamjarling, line 3: Dig the `use-iiif` title...clean.

src/components/Viewer/Viewer.tsx
```diff
@@ -73,6 +70,7 @@ const Viewer: React.FC<ViewerProps> = ({ manifest, theme }) => {
       activeCanvas,
       "text/vtt",
     );
+    console.log(resources);
```
adamjarling, line 73: Do we need this info?

adamjarling (approved):
Looking good.

PR #78, opened 2022-04-21
Upgrade to React 18

adamjarling:
Upgrade to React 18; make esbuild target lower to es2019 for more forgiving consuming environments.  This allows consuming environments to be running React v 16.13 through v 18 and importing Clover won't crash consuming apps.


mathewjordan (approved):
🚀 🤣 Good to go.

PR #83, opened 2022-04-25
Fix buggy OSD in React v18.

mathewjordan:
- Refine OSD instancing, fix undefined renders.
- Update tests.
- Remove overflow, target class selectors rather than ID.


adamjarling (approved):
Looks good to me.

PR #84, opened 2022-04-28
Add config options for active canvas and styling flexibility.

mathewjordan:
- Add basic classes for selector styling.
- More flexible styling for consuming app containers.

This adds two new config options for managing height and background color of the active canvas. These values are fed into the viewer context and distilled down to the video player and OpenSeadragon for respective content resources types. 

image

I also refreshed some styling choices that made to make Clover much more flexible in consuming applications.
- Added basic `clover-*` class names to some high level layout elements
- Removed padding wrapping entire app so that Clover reaches the edges of the container in the consuming application
- Replaced `<section>` wrapping element with a generic `<div>`
- **and** Cleaned up some responsive behavior on the media thumbnails

### DIFF
--- a/src/components/ImageViewer/OSD.tsx
+++ b/src/components/ImageViewer/OSD.tsx
@@ -4,6 +4,7 @@ import { Navigator, Viewport, Wrapper } from "./ImageViewer.styled";
 import Controls from "./Controls";
 import { getInfoResponse } from "services/iiif";
 import { v4 as uuidv4 } from "uuid";
+import { useViewerState } from "context/viewer-context";
 
 export type osdImageTypes = "tiledImage" | "simpleImage" | undefined;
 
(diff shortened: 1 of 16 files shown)


src/components/ImageViewer/OSD.tsx
```diff
@@ -14,6 +15,9 @@ interface OSDProps {
 
 const OSD: React.FC<OSDProps> = ({ uri, imageType }) => {
   const [osdUri, setOsdUri] = useState<string>();
+  const viewerState: any = useViewerState();
```
adamjarling, line 18: Could we easily pull in viewer state `type` here?   If not no biggie.

src/components/Viewer/Content.tsx
```diff
@@ -36,7 +36,7 @@ const ViewerContent: React.FC<Props> = ({
   isNavigatorOpen,
 }) => {
   return (
-    <Content>
+    <Content className="clover-content">
```
adamjarling, line 39: Nice and helpful

src/components/Viewer/Content.tsx
```diff
@@ -53,7 +53,7 @@ const ViewerContent: React.FC<Props> = ({
             </Button>
           </CollapsibleTrigger>
           {items.length > 1 && (
-            <MediaWrapper>
+            <MediaWrapper className="clover-canvases">
```
adamjarling, line 56: Same, very helpful

adamjarling (approved):
Looks great.

PR #87, opened 2022-08-01
Allow setting of OpenSeadragon Options, default scrollToZoom as false.

mathewjordan:
## What does this do?

This sets the default **scrollToZoom** functionality as `false` for the OpenSeadragon options. Along with this, I have included the ability for a Clover implementer to modify (with some possibility for danger) the OpenSeadragon options through the **configOptions.openSeadragon** prop.

**What doesn't this do?**
Well, the cursor can still be _captured_ currently by OpenSeadragon canvas. I believe we might need to dig into that in a separate issue as it is somewhat annoying if you're trying to scroll past the canvas and are not aware that your cursor should be outside of the bound of it to affect the body scroll.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 **A minimal IIIF viewer for Image, Audio, and Video canvases built with React.js**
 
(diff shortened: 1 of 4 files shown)


adamjarling (approved):
Looking good.

PR #88, opened 2022-09-07
Add iiif-content param capability to demo page.

mathewjordan:
## What does this do?

This adds functionality to read a `iiif-content` URL parameter and set it as the `manifestId` prop for Clover on the demo/development view. Doing this should bring us a step closer to Clover being included in the **iiif.io**[ viewer matrix](https://iiif.io/api/cookbook/recipe/matrix/).

**Note** -- This does not yet accommodate Clover for [Content State Encoding](https://iiif.io/api/content-state/0.9/#6-content-state-encoding).

image

### DIFF
--- a/src/dev.tsx
+++ b/src/dev.tsx
@@ -1,4 +1,4 @@
-import React, { StrictMode } from "react";
+import React, { StrictMode, useEffect, useState } from "react";
 import App from "@/index";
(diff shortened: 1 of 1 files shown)


adamjarling (approved):
Looks easy enough

PR #89, opened 2022-09-07
Support IIIF Collections

mathewjordan:
## What does this do?

This work brings Clover to a point where it can support IIIF resources of type `Collection` by use of a `id` prop. 

image

If  IIIF resource is a **Collection**, a Radix UI `Select` will render as the title of current manifest. By default, at this point, the first item in the Collection `items` will render as the **activeManifest**. If a user interacts with the select dropdown, they will see a list of the Manifest  items. Selecting one will dispatch a new activeManifest and in turn send that manifest and its initial canvas into the `Viewer` component.

One key update is that the IIIF resource prop should now be `id` rather than `manifestId`. `manifestId` will still be supported by is no longer required. In a future version we may want fully deprecate this.

## How to review?

This adds one new  Radix UI package, so do `npm ci` to install that.


mathewjordan:
> @mathewjordan just wanted to add that I threw a bunch of our collections at this this morning. Overall, things work well, but it definitely struggles with "large" collections. Take for instance [Roth](https://digital.lib.utk.edu/assemble/collection/gsmrc/roth). This collection has a little over 7000 works. If I throw this at Mirador, it takes a few seconds to load the data (5 seconds?), but it takes much longer in Clover to get just the items to load (at least a minute). After that, it takes another minutes or so for me to even be able to interact with the list of works. Eventually (several minutes), thumbnails and items are navigable, but I've yet to get an item to load after this initial screen:
> 
> image
> 
> I think my question is the goal. Is the intent for this to work for collections of any size or simply have some basic functionality with collections. If it's the latter, I think this works well, but I question the limits of a collection this can support.
> 
> I also wonder if our stack might be causing additional issues with performance, but again this seems to work okay in Mirador. Of course, Mirador goes for a much more basic approach to loading these things.
> 
> I'm going to continue testing to figure out what the collection size looks like when things go from slow to unusable.

For very large Collections, the best method may be to lazyLoad. Since we have already have similar issues with performance on thumbnails for canvases in large manifests, I think it'd be best to tackle lazyLoad as one larger issue rather than within this PR.

src/index.tsx
```diff
@@ -65,34 +72,76 @@ const RenderViewer: React.FC<Props> = ({
...
+        .finally(() => {
+          dispatch({
+            type: "updateIsLoaded",
+            isLoaded: true,
+          });
+        });
+  }, [activeManifest]);
+
+  useEffect(() => {
+    let resource = id;
```
adamjarling, line 97: Could you do `const resource = id ? id : manifestId`, or something like that maybe?

adamjarling (approved):
This looks great.  Super solid work.

PR #92, opened 2022-09-19
Use Nectar Label component where possible.

mathewjordan:
This is just a quick PR to make a minor update to Clover, using Nectar `<Label>` wherever a IIIF [label](https://iiif.io/api/presentation/3.0/#label) is rendered.

- Update manifest label.
- Replace rendered labels with Nectar component.
- Use getLabel on alt.


adamjarling (approved):
Definitely cleans up the code.  And best branch name ever, "nectarize".

PR #93, opened 2022-09-20
Create manifest information panel for IIIF properties.

mathewjordan:
## What does this do?

This work adds an **About** tab that renders in the Navigator `aside` by default. Boolean values for `renderAbout` and `showInformationToggle` can now be set in the `options` prop.

Boolean values within the `options` prop, `renderAbout` and `showInformationToggle`, relate to rendering Manifest content in an `<aside>` and providing user ability to close that panel. The toggle functionality is handled with a [Radix UI Switch](https://www.radix-ui.com/docs/primitives/components/switch).

The **About** will render Manifest level properties via [Nectar](https://github.com/samvera-labs/nectar-iiif) for:
- `summary` as a `<p>`
- `metadata` as a `<dl>`
- `requiredStatement` as a`<dl>`
- `rights` as an `<a>`
- `homepage` as a `<ul>`
- `seeAlso` as a `<ul>`
- `id` as an `<a>`

image

## IIIF Recipe Support
This addition should bring Clover to support the following recipes:

- [Internationalization and Multi-language Values (label, summary, metadata, requiredStatement)](https://iiif.io/api/cookbook/recipe/0006-text-language/)
- [Embedding HTML in descriptive properties (label, summary, metadata, requiredStatement)](https://iiif.io/api/cookbook/recipe/0007-string-formats/)
- [Metadata on any Resource (metadata)](https://iiif.io/api/cookbook/recipe/0029-metadata-anywhere/)
- [Linking to Structured Metadata (seeAlso)](https://iiif.io/api/cookbook/recipe/0053-seeAlso/)
- [Image Thumbnail for Manifest (thumbnail)](https://iiif.io/api/cookbook/recipe/0117-add-image-thumbnail/)

And potentially supports:

- [Displaying Multiple Values with Language Maps (label, summary, metadata, requiredStatement)](https://iiif.io/api/cookbook/recipe/0118_multivalue/)

## Things not yet fully considered
- Multi-language support for the hardcoded *About*  tab title and *More Information* toggle
- Responsiveness of Navigator on `sm` and lower viewerports

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,11 +67,9 @@ return <CloverIIIF id={id} />;
 
 [See Example](https://codesandbox.io/s/samvera-clover-iiif-i0huq)
 
-
 <h3>Usage with Next.js</h3>
 
 Usage with Next.js requires a dynamic import using `next/dynamic`due to a depdency of OpenSeadragon.
-
 
 ```jsx
 import dynamic from "next/dynamic";
@@ -82,9 +80,7 @@ const CloverIIIF = dynamic(() => import("@samvera/clover-iiif"), {
 
 const id = "https://digital.lib.utk.edu/assemble/manifest/heilman/1187";
 
-
-return <CloverIIIF id={id} />
-
+return <CloverIIIF id={id} />;
 ```
 
 ---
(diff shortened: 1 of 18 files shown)


markpbaggett (approved):
@mathewjordan sorry for the delay.  I think this looks good.  

A couple of oddities:  I don't understand why `seeAlso` links have trouble with label. They work fine in pure nectar, but not here:  

image

I don't think this matters, but I think we should put an issue to look into this post merge.

Also, as we discussed, I think there are 2 other things that we need to work on post merge.

1. Customization and / or Internationallization of the "About" / "More Details" values.
2. Determine how to know which language to display for  metadata elements in "About." Personally, I think a good first step would just be to make nectar display all language values by default unless a specific bcp-47 code is provided and then tie that updated nectar in here.

Regardless, I think this is good enough for now as is to be merged.

PR #94, opened 2022-09-23
Address seeAlso label rendering issue.

mathewjordan:
## What does this do?

This resolve an issue where the `id` as the innerHTML within the `<a>...</a>` for each `seeAlso` entry. The issue related to how `vault` stores resources locally by **id**, I am no making `vault.get()` request to the Redux store and grabbing the additional data for entries.

Clover uses Nectar for rendering a seeAlso, thus it will still fallback to rendering an `id` if the `label` does not exist.

I also made `homepage` render in similar way, and fixed a small typo related issue with `manifest.id` not rendering.


markpbaggett (approved):
@mathewjordan Awesome! Looks good.  😎

PR #95, opened 2022-09-24
Breakout IIIF properties, add tests.

mathewjordan:
## What does this do?

This is some followup work on #94 that breaks out the various IIIF properties rendered in `components/Navigator/About/About` into individualized components. By doing so, we can easily test the expected output of these. Baseline unit tests for each property are now included.

This also adds this `jest-environment-jsdom` package, so make sure to do `npm ci` or `npm i`.

### Note

Tests included here are for Clover specifically and thus should be simple rendering tests for how Clover chooses to render layout. More complex testing on varying ways for a Client to render IIIF properties should be handled in [Nectar ](https://samvera-labs.github.io/nectar-iiif/).

## To see in action...

Run `npm run test src/components/Properties` to watch.

image


markpbaggett (changes requested):
@mathewjordan This looks like a good initial way to bring in Jest, but it doesn't add `jest-environment-jsdom`. Because of this, when I run tests, I get this message:


```
As of Jest 28 "jsdom" is no longer shipped by default, make sure to install it separately.
```

To fix this, I had to add `jest-environment-jsdom`.  Is there a reason why we shouldn't do this?

mathewjordan:
> @mathewjordan This looks like a good initial way to bring in Jest, but it doesn't add `jest-environment-jsdom`. Because of this, when I run tests, I get this message:
> 
> ```
> As of Jest 28 "jsdom" is no longer shipped by default, make sure to install it separately.
> ```
> 
> To fix this, I had to add `jest-environment-jsdom`. Is there a reason why we shouldn't do this?

Yup, I noticed the same thing. `jest-environment-jsdom` is included as a package now. Sorry, I omitted explicit install instructions. Check the package.json.

markpbaggett (approved):
Looks good.  :shipit:

PR #101, opened 2022-12-13
Spread out openSeadragon options on OpenSeadragon instances (allows for ajaxWithCredentials)

mathewjordan:
Spreads the already existing and seemingly unused `openSeadragon` property out after defaults. This will allow a user to customize their OpenSeadragon instance more fully using https://openseadragon.github.io/docs/OpenSeadragon.html#.Options.

`options.openSeadragon` now has a default of `{}` and is better documented.

Example usage:
```jsx
      <CloverIIIF
        id={url}
        options={{
          openSeadragon: {
            ajaxWithCredentials: true,
          },
        }}
      />
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,20 +200,21 @@ return <CloverIIIF id={id} customTheme={customTheme} />;
 
 <h2 id="reference">Reference</h2>
 
-| Prop                            | Type       | Required | Default   |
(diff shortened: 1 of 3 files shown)


mbklein (approved):

adamjarling (approved):
I like it.  We had to do something similar with the React OpenSeadragon Viewer too...

PR #102, opened 2022-12-14
Add custom fetcher and withCredentials.

mathewjordan:
This adds a custom fetcher to our Vault instances, and allows for the boolean option `withCredentials` to be distilled down to the `customFetcher` using axios.get().

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@samvera/clover-iiif",
-  "version": "1.12.5",
+  "version": "1.12.6-rc.0",
   "description": "Viewer for audio, video and image file types driven by a IIIF manifest",
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -70,6 +70,7 @@
     "@radix-ui/react-tabs": "^1.0.1",
     "@samvera/nectar-iiif": "^0.0.18",
     "@stitches/react": "^1.2.8",
+    "axios": "^1.2.1",
```
adamjarling, line 73: Do we have to use `axios`?

adamjarling:
@mathewjordan This looks fine, but if there's an opportunity to keep this light (no axios or anything else there's native ways of doing), it'd be nicer for consumers and maybe for us too down the road.

mathewjordan:
@adamjarling Good call. For community stuff shying away from axios might be best. Instead of opting towards `fetch`, I decided to construct a small XHR get service. I'm doing this to be congruent with `withCredentials` (https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) as fetch's `credentials` (https://developer.mozilla.org/en-US/docs/Web/API/fetch#credentials) is an enum and not exactly on/off as we'd probably want this to be. This will also match our use of `withCredentials` elsewhere in our consuming apps like DC.

kdid (approved):

PR #111, opened 2023-03-09
Make OSD controls more usable in smallest viewports.

mathewjordan:
## What does this do?

This adds a small styling responsive fix to orient the OpenSeadragon controls vertically at `xs` viewports. The zIndex of the Controls is also now at a higher level than the "Navigator" in the top left, assuring that the controls are always going to be clickable.


adamjarling (approved):
Very usable!

PR #113, opened 2023-03-16
Support IIIF Presentation API 3.0 `placeholderCanvas`

mathewjordan:
## What does this do?

This brings support to Clover for IIIF Presentation API 3.0  property `placeholderCanvas` for `Canvas` resources. If a placeholderCanvas exists, the annotation body of that canvas will render in place of the Canvas. An button with an icon will display to the user indicating that clicking on the placeholderCanvas OR its toggle button will open the respective Canvas in either OpenSeadragon (Images) or a media player (Video, Sound). At this point users will be in a state of `isInteractive` being true. This will stay true until the user hits the close button toggle -- even if the user toggles through canvases on the Manifest.

### Placeholder Canvas on a multi-canvas Manifest (Image)

image

### Placeholder Canvas on a Manifest (Video)

image

### Notes
- Image placeholders have a _zoom in_ icon for open, while Video and Sound will have a _play_ button.
- Video and Sound will render a toggle button that centers on the screen
- Video and Sound canvases with a placeholder will automatically play onClick

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@iiif/vault": "^0.9.19",
+        "@iiif/vault-helpers": "^0.9.11",
         "@nulib/design-system": "^1.6.0",
         "@radix-ui/react-collapsible": "^1.0.1",
         "@radix-ui/react-radio-group": "^1.1.0",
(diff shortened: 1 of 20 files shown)


adamjarling (approved):
This is fantastic.  I like keeping all the nav controls consistent top right, and nice work calling attention to the exit control by color contrast.  Feels great.  

The only small thing I'm noticing, is maybe adding a `cursor: pointer` style when hovering over the Magnifying Glass icon when not yet engaged in "interactive" mode.   Just so it's not magnifying a magnifying feel?  Everything else is so slick, this could be icing on the cake, but  up to you.

Great work!

PR #115, opened 2023-04-07
Fix zIndex issue on audio (sound) canvases.

mathewjordan:
## What does this do?

This fixes an issue on Audio canvases where the visualizer canvas has overlaps the the `<video>` element playing the canvas. zIndex properties have been adjusted to ensure the audio visualizer renders below the player which has a transparent background.


bmquinn (approved):
Looks great! 👍

PR #116, opened 2023-04-14
Improve process and rendering of long WebVTT resources for navigator.

mathewjordan:
## What does this do?

This was a doozy.

Two primary issues are resolved in this work.

- Wrapping Context is removed that tracked time and is replaced by an event listener in the `Cue` tracking currentTime of the `<video>` element. Previous to this work, all cues were constantly being rerendered every time time was updated on the video. In this method, Cue items will only update if the cue's `isActive` status changes.
- The `createNestedCues` function has been rewritten to be more performant (with the help of AI). Previously, this really dragged the intiial load of long WebVTTs.

Not covered:

- I opted against a lazy load of the WebVTT cues for the time being. Performance improved dramatically once the two items above were resolved. If we deem it important, we can work that in at a later date.

## How should we review?

- The best way to review this is to test using the Manifests with Video that also have a WebVTT. There are two included in the Manifests tray in dev mode. 
- Are WebVTTs still nested? Are they interactive? How bad is the initial load (previously it was 20 seconds on the UTexas one) following by a complete lack of usability in interaction.


adamjarling (approved):
Looks good.  The `createNestedCues()` function is zippy, and the lag looks like just rendering out the list.  Agreed @mathewjordan if there's issues, might not hurt to test run [Virtualized Lists](https://github.com/bvaughn/react-virtualized) or something equivalent, to benchmark it?

Good work!

PR #119, opened 2023-05-19
Update `poster` attribute on A/V canvases.

mathewjordan:
## What does this do?

This updates our `<Player />` component for A/V canvases to now utilize any `placeholderCanvas` and/or `accompanyingCanvas` properties as the value for the `poster` attribute. There is potential for these to conflict and some logic has been built in to check for currentTime on the video element. If a conflict exists, the placeholder is used ONLY of if the currentTime is `0`.

## How to review?

The dev instance has three different Manifests to test from. Each of these should render a poster in the Video element. The first one, _accompanying/placeholder canvas_, will render the placeholderCanvas initially and then accompanyingCanvas while playing.

- accompanying/placeholder canvas
- accompanyingCanvas
- placeholderCanvas

Image placeholderCanvases should still function as previously designed with a "placeholder" image rendered initially prior to the OpenSeadragon component being rendered onClick, test w/ https://dcapi.rdc-staging.library.northwestern.edu/api/v2/works/fc73b3ed-2e13-4ec7-9bff-dd540563d7c9?as=iiif

### DIFF
--- a/src/components/Player/Player.tsx
+++ b/src/components/Player/Player.tsx
@@ -1,4 +1,4 @@
-import React, { MutableRefObject, useEffect } from "react";
+import React, { useEffect } from "react";
 import Hls, { HlsConfig } from "hls.js";
 import { PlayerWrapper } from "@/components/Player/Player.styled";
 import { IIIFExternalWebResource } from "@iiif/presentation-3";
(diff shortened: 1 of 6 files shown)


src/components/Player/Player.tsx
```diff
@@ -113,11 +93,37 @@ const Player: React.FC<PlayerProps> = ({
...
+    const canvas = vault.get(activeCanvas);
+    const accompanyingCanvas = canvas.accompanyingCanvas?.id
+      ? getPaintingResource(vault, canvas.accompanyingCanvas?.id)
+      : null;
+
+    const placeholderCanvas = canvas.placeholderCanvas?.id
+      ? getPaintingResource(vault, canvas.placeholderCanvas?.id)
+      : null;
+
+    const conflictingCanvas = !!(accompanyingCanvas && placeholderCanvas);
```
adamjarling, line 105: 👍

adamjarling (approved):
Looks good.  Nice, clean implementation.

Issue #122, opened 2023-07-27
development server continuously hot reloading


wykhuh:
I cloned this repo, installed the packages, and started the development server. The server is continuously hot reloading, which causes the browser to also continuously reload.  

If I change `server.js`  `.on("all"`  to `.on("change" `,  then the server only reloads if I make a change to a file. Is there a reason to use  `.on("all"` ?

mathewjordan:
@wykhuh You are correct. I don't think we need `all`. 

Would you like me to make this change? Or will you put in a pull request?

Thank you for your interest in the project!

wykhuh:
@mathewjordan I made a pull request. https://github.com/samvera-labs/clover-iiif/pull/123

I'm a developer for [Collective Access](https://collectiveaccess.org), an open source cataloguing software used by museums, archives, and other groups. We rewriting our frontend app, and we are considering using Clover as our media viewer. Our users have IIIF images, video, audio, pdfs, and 3D models.

mathewjordan:
> @mathewjordan I made a pull request. #123
> 
> I'm a developer for [Collective Access](https://collectiveaccess.org), an open source cataloguing software used by museums, archives, and other groups. We rewriting our frontend app, and we are considering using Clover as our media viewer. Our users have IIIF images, video, audio, pdfs, and 3D models.

Thank you put the PR through @wykhuh. I really appreciate that.

If you're interested in implement Clover, you may want to know that I'm currently reorganizing Clover in advance of a version 2.x release. Clover IIIF will now be a suite of tools (with the `Viewer` just being one component) aimed at composing IIIF-centric user interfaces. If you'd like to take a look at that, please see docs in development here https://mathewjordan.github.io/clover-iiif-proto/ and the [repo](https://github.com/mathewjordan/clover-iiif-proto).

This will allow developers to quickly create interfaces like https://dc.library.northwestern.edu/items/8128dc86-9214-49a5-b40c-3b6416dfe740 using simple React components.

PR #123, opened 2023-07-27
fix live reload on dev server

wykhuh:
The dev server is continuously reloading because watch mode is set to `all`. This PR changes the watch to `change`, so it only reloads when there is a change in the file.  Related issue #122


mathewjordan (approved):
👍

PR #124, opened 2023-08-01
Add Primitives Provider

adamjarling:
## What does this do?
Adds in the `Primitives` Context/Provider which we wrap `Primitives` components in if they need to support extra user configuration, ie. `customValueDelimiter`.


mathewjordan (approved):
Looks great. Should we get tests integrated with GitHub actions as well?

adamjarling:
@mathewjordan Yea might as well get the Github Actions Tests wired up.  I'll push up a commit in a few...

adamjarling:
@mathewjordan This is now set up to run Tests against all PRs to the `suite` branch.  Configured for `pnpm` as well, which we'll need to update the `.github/workflows/gh-pages.deploy.yml` at some point too when we get close to merging with `main`.

PR #129, opened 2023-08-03
Update `Slider` docs.

mathewjordan:
This adds documentation for the `Slider` component,. fixes a naming issue in the Viewer docs, and removes an unused prop.


adamjarling (approved):
Yep, good to get in there so I can tweak stuff if need be.

PR #131, opened 2023-08-03
Slider test coverage partial PR

adamjarling:
## What does this do?
Partial PR for starting to cover `Slider` components with Unit Tests.   Figure it'll be easier incrementally than saving for one big one.

## How to test
``` bash
pnpm test
``` 

should all pass.


mathewjordan (approved):
Love it, thanks for taking on adding the coverage.

PR #132, opened 2023-08-03
Updata Metadata docs.

mathewjordan:
This updates the Metadata docs.

---

I did this as-is for now but noticed an issue with how `metadata` outputs to `dl`, `dt`, `dd` elements that we may want to fix.

Issue is that each individual value entry does not have its own `dd` element. I think if we should probably fix this if we want to stick to semantic HTML as much as possible. This definitely is related to the customDelimiter.

**Current**

```html
<dl>
  <div role="group" data-label="subject">
    <dt>Subject</dt>
    <dd>Masks, Commedia dell'arte,  Italian drama (Comedy),  Zanni (Fictitious character)</dd>
  </div>
</dl>
```
**Proposed**
```html
<dl>
  <div role="group" data-label="subject">
    <dt>Subject</dt>
    <dd>Masks</dd>
    <dd>Commedia dell'arte</dd>
    <dd>Italian drama (Comedy)</dd>
    <dd>Zanni (Fictitious character)</dd>
  </div>
</dl>
```

### DIFF
--- a/pages/docs/label.mdx
+++ b/pages/docs/label.mdx
@@ -114,7 +114,7 @@ The `Label` component can be rendered as ` span`, `h1`, `h2`, `h3`, `h4`, `h5`, 
   </Tab>
   <Tab>
     ```html
-    <span class="c-PJLV custom-label-class" id="custom-label-id">
+    <span class="c-PJLV custom-label-class">
       Self-Portrait Dedicated to Paul Gauguin
     </span>
     ```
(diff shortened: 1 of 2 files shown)


adamjarling:
@mathewjordan Yea, the `<dl> <dt> <dd>` makes sense.  So we'd need to handle that via CSS?   Making the `dd`'s inline or something, then popping in the custom delimiter after each `dd`?  Good catch.

mathewjordan:
> @mathewjordan Yea, the `<dl> <dt> <dd>` makes sense. So we'd need to handle that via CSS? Making the `dd`'s inline or something, then popping in the custom delimiter after each `dd`? Good catch.

I think some sort of `::after` for each `dd` could contain the delimiter. Maybe the default delimiter should be `undefined`?

PR #137, opened 2023-08-23
Remove @nulib/design-system as a dependency

adamjarling:
## What does this do?
- Removes `@nulib/design-system` as a package dependency
- Copies over the previously imported components to `src/components/internal` directory
- Update the Stitches CSS config a bit
- Remove `Button` component, creates new versions of the rest.

I think this will be a cleaner implementation via the eyes of general public, even though we are extending our dependency on Stitches.dev, but we'll deal with that all in one go I imagine. 

## To Test
Start up the app and things should run and look just like before.


mathewjordan (approved):
Glad we are shedding that. Should make future styling updates a little more straight-forward too. This lead me down road of breaking out Shared components across Viewer/Slider/etc.. at some point too?

adamjarling:
> Glad we are shedding that. Should make future styling updates a little more straight-forward too. This lead me down road of breaking out Shared components across Viewer/Slider/etc.. at some point too?

@mathewjordan Yep, we're on the same page... figured establish a small pattern here, and I'll create a new issue to account for further isolation/breakdown of Shared components.  👍

PR #138, opened 2023-08-23
Support custom request headers, including `"Authorization": "Bearer ${token}"`

mathewjordan:
## What does this do?

This adds the ability for customization of the request headers made by the Clover viewer. For example:

Noted in the revised documentation for the VIewer component:

> In some cases, a client may need to request Manifest or Collection resources with custom request headers, ex: `Authorization`. This can be done by passing a `requestHeaders` object to the `options` prop. This object will be passed to the request call made by the Viewer. Accepted **header** keys are defined in the [IncomingHttpHeaders](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/http.d.ts) interface.

Example:

```jsx
const iiifContent =
  "https://api.dc.library.northwestern.edu/api/v2/works/8a833741-74a8-40dc-bd1d-c416a3b1bb38?as=iiif";

const token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9...";

return (
  <Viewer
    iiifContent={iiifContent}
    options={{
      requestHeaders: {
        "Content-Type": "application/json",
        Authorization: `Bearer ${token}`,
      },
    }}
  />
);
```

Additionally, adjustments were made to allow Viewer navigator component Cues and Tabs to inherit color from the consuming application. The background color of the canvas is also less stark now by using alpha transparency.

### DIFF
--- a/src/lib/xhr.ts
+++ b/src/lib/xhr.ts
@@ -1,6 +1,8 @@
+import { IncomingHttpHeaders } from "http";
+
 export interface RequestOptions {
   ignoreCache?: boolean;
-  headers?: { [key: string]: string };
+  headers?: IncomingHttpHeaders;
   timeout?: number;
   withCredentials: boolean;
 }
@@ -36,7 +38,7 @@ function parseXHRResult(xhr: XMLHttpRequest): RequestResult {
 
 function errorResponse(
   xhr: XMLHttpRequest,
-  message: string | null = null,
+  message: string | null = null
 ): RequestResult {
   return {
     ok: false,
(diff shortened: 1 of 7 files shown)


src/lib/xhr.ts
```diff
@@ -1,6 +1,8 @@
+import { IncomingHttpHeaders } from "http";
+
 export interface RequestOptions {
   ignoreCache?: boolean;
-  headers?: { [key: string]: string };
+  headers?: IncomingHttpHeaders;
```
adamjarling, line 5: 👍

adamjarling (approved):
Looks great.  Nice, simple, clean implementation.

PR #143, opened 2023-09-05
4164 options information open

mathewjordan:
- Update higher level options.
- Drilldown on information panel open.


adamjarling (approved):
Nice clean up, good new naming conventions, looks solid.

PR #144, opened 2023-09-12
Add non-next dev ability. Address accessibility concerns.

mathewjordan:
## What does this do?

1. This adds the ability for us to test Clover components in a vite-spun environment without Nextra. This can be run by: `npm run dev:vite`.
2. Uses Lighthouse to assess accessibility of components, initial score be **82**.

image

3. Address issues:
- Name issue on information toggle (resolved with combination of arial-label and name attributes)
- Contrast issue on Slider header elements (resolved by combination of allowing inheriting header text color, and making our default accent color a bit darker)
- The remaining `dl` issue(s) are not actually a modern problem. It looks like Lighthouse uses an outdated spec in its analysis and does not account for [this ~2016 change](https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element) that allows `div` groups of `dt` and `dd` values within a `dl`.  
- Given this, I think the 96 is actually a 💯 

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,6 +8,6 @@
   </head>
   <body>
     <div id="root"></div>
-    <script type="module" src="./index.tsx"></script>
+    <script type="module" src="/src/dev.tsx"></script>
   </body>
 </html>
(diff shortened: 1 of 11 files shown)


adamjarling (approved):
Nice addition!  Runs as expected and great news on the 96/100 however it scores...excellent.

PR #148, opened 2023-10-10
Support `Choice` annotations.

mathewjordan:
## What does this do?

This is a big one...

This adds the ability for a user to select between options a `Choice` annotation type.
https://www.w3.org/TR/annotation-model/#choice-between-bodies

image

If a Choice exists on the active canvas, the options are presented to the user in a select render below the visible canvas (OpenSeadragon for Image, HTML5 video for Video and Sound). The work he abstracts out the Collection select and makes this component generic under `src/components/internal/Select`. The newly created Select component is then used for both the Collection drop down in the header, functioning as it did previously, and as the Choice select UI element. In the case of Choice, the the default option is the first item in the Choice array.

When reviewing, the following manifests should accomplish different things:

1. An image manifest with choice will render the `ImageViewer` component with a Choice select below it. With this, Clover now supports the [Multiple Choice of Images in a Single View (Canvas) recipe](https://iiif.io/api/cookbook/recipe/0033-choice/) see: [localhost:3000/docs/viewer/demo?iiif-content=https://iiif.io/api/cookbook/recipe/0033-choice/manifest.json](localhost:3000/docs/viewer/demo?iiif-content=https://iiif.io/api/cookbook/recipe/0033-choice/manifest.json) locally.

2.  A video manifest with choice will render the `Player` component with a Choice select below it. With this, Clover will now support Avalon and Internet Archive manifests with choice options. See: [localhost:3000/docs/viewer/demo?iiif-content=http://localhost:3000/manifest/avalon/transcript-annotation.json](localhost:3000/docs/viewer/demo?iiif-content=http://localhost:3000/manifest/avalon/transcript-annotation.json) and [localhost:3000/docs/viewer/demo?iiif-content=https://iiif.archive.org/iiif/Shogun_Miniseries/manifest.json](localhost:3000/docs/viewer/demo?iiif-content=https://iiif.archive.org/iiif/Shogun_Miniseries/manifest.json) locally. Note that the Internet Archive video formats may not currently be support by the browsers native HTML5 video element. It seems like **Ogg Video** is commonly supported.

3. Collections should still function as expected, with a select render in the Header in lieu of a static text Manifest label. See: [http://localhost:3000/docs/viewer/demo?iiif-content=https://canopy-iiif.github.io/canopy-iiif/api/facet/date/1910.json](http://localhost:3000/docs/viewer/demo?iiif-content=https://canopy-iiif.github.io/canopy-iiif/api/facet/date/1910.json) locally.


src/types/presentation-3.ts
```diff
@@ -0,0 +1,9 @@
+import {
+  IIIFExternalWebResource,
+  InternationalString,
+} from "@iiif/presentation-3";
+
+export interface LabeledIIIFExternalWebResource
```
adamjarling, line 6: I like this pattern of extending out some of the Presentation 3 package types.

adamjarling (approved):
Looks and performs well.  Lots of code updates 👍

PR #149, opened 2023-10-22
Custom display proof of concept

adamjarling:
## What does this do?
Allows Clover's `Viewer` component to bypass default "viewer components" (Open Seadragon and HTML Video player) and render custom display component(s).  A consuming application would pass in a custom React component to `Viewer`, alongside targeting criteria for which items in the source IIIF Manifest should be rendered by the custom display.

### Motivation
Motivation for this POC is from a use case proposed needing to display a Collection of PDFs using `pdf.js`.  

### How does it work?
Create your own "viewer" component.

Then configure the `Viewer` to target Manifest items in one of two ways:
1. Target by Canvas `id`s, or
2. Target by a `mime/type` value (ie. `application/pdf`), which is represented by the `painting.format` property in Clover's Viewer (see `src/components/Viewer/Painting/Painting.tsx`)

```tsx
import { Viewer } from "@samvera/clover-iiif";

function MyDisplayComponent({ id }) {
  return <div>...</div>;
}

type Target = {
  target: {
    canvasIds?: string[],
    paintingFormat?: string
  }
}

<Viewer
  iiifContent="https://manifest-url-here"
  customDisplays={[
    {
      target: {
        // Target by canvas ids
        canvasIds: ["http://a-canvas-id-in-manifest/access/0"],
        // Or by paintingFormat
        paintingFormat: "image/tiff" // "application/pdf", etc...
      },
      component: MyDisplayComponent,
    },
  ]}
/>
```

A user defines their own custom React component, which wraps their own display code/logic/parsing/etc.  Clover's `Viewer` passes back an `id` prop to the custom component, which is the Canvas `target` value . The custom component (in this situation) is then responsible for parsing the manifest, finding the source item to display, and then free to render the source content however they like.

```tsx
// Clover's Viewer will pass back the Canvas id used above in "customDisplays" prop
const MyDisplayComponent = ({ id }) => {
  // Some psuedo code
  const canvas = getCanvas(id);
  const src = getSourceOfTheItem(canvas.items[0].items[0].body);
  return <div><img src={src} alt={canvas.label} /></div>;
}
```
For a more complete example, see this file in the PR:  `docs/components/CustomDisplays/PDFViewer.tsx`

## How to test?
1. Run the branch locally
3. Navigate to the "Viewer Demo" page, it will default to using a local manifest which contains a PDF fixture as the first of three Canvases in the Manifest file (`public/manifest/custom-displays/pdf.json`).  The PDF file has a `placeholderCanvas` defined, and will follow the `Viewer` pattern of "clicking into" a Work.
4. In the Manifest URL bar, test out a second fixture manifest (`public/manifest/custom-displays/pdf-no-placeholder.json`).  This example has no `placeholderCanvas` and might be another example of how a user would want to implement.

## More Notes
@mathewjordan I'm not sure we want to go down the road of passing back `painting` as an expected prop to user-defined Custom Display components.  I made some comments in the code.  It is a nice convenience, but maybe it's a better pattern to keep maximum control on the client-side for flexible implementations?


mathewjordan:
This one ready for review @adamjarling ?

adamjarling:
@mathewjordan @davidschober Yea, I think it's definitely ready for review but I'm not sure it's a priority to merge?  Why don't you put some eyes on it Mat, we can further refine a Draft PR, and then run it by Collective Access to see how this looks from their end?

I'm also writing some tests for the `Painting` component (and it's underlying components), so at least a commit, or a few testing commits from this branch could be cherry-picked on another branch, independent of "Custom Display" code, which could be ready to merge earlier.

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "editor.formatOnSave": true,
   "editor.defaultFormatter": "esbenp.prettier-vscode"
```
mathewjordan, line 6: Should we have this file in the repo?

pages/docs/viewer.mdx
```diff
@@ -320,3 +332,39 @@ return (
...
+
+```tsx
+import MyCustomDisplay from './MyCustomDisplay';
+import AnotherCustomDisplay from './AnotherCustomDisplay';
+
+type CustomDisplay = {
+  component: React.ElementType;
+  target: {
+    canvasId: string[];
+    paintingFormat: string[]; // "application/pdf" or "application/epub+zip"
```
mathewjordan, line 350: This declares `paintingFormat` as an array of strings, but in the below example, it is a string. I think an array is solid and will provide the most flexibility.

src/components/Viewer/Painting/Painting.tsx
```diff
@@ -36,11 +37,26 @@ const Painting: React.FC<PaintingProps> = ({
...
   };
 
+  const customDisplay = customDisplays.find((display) => {
+    let match = false;
+    const { canvasId, paintingFormat } = display.target;
+    if (canvasId) {
+      match = canvasId.includes(activeCanvas);
+    }
+    if (paintingFormat) {
+      match = paintingFormat === painting[annotationIndex]?.format;
```
mathewjordan, line 52: Going with the array notion above, this would need to be some form of `includes()`

pages/docs/viewer.mdx
```diff
@@ -320,3 +332,39 @@ return (
   />
 );
 ```
+
+---
+
+### Custom displays
```
mathewjordan, line 338:
```suggestion
### Custom canvas displays
```

I wonder if we should use the word `canvas` or `Canvas` here to speak towards a IIIFy way of thinking? The canvas is what the annotation resource is _painted_ on.

mathewjordan (changes requested):
Just a few notes. This is solid and should provide exceptional flexibility. The one thing I think we should absolutely do is make `paintingFormat` an array, so a user can apply a custom display to multiple formats, ex:

```json
["audio/ogg", "video/ogg"]
```

.vscode/settings.json
adamjarling: Nope, we should not.  Good catch 👍

pages/docs/viewer.mdx
adamjarling: Yea, I like being specific here.  Good call.

adamjarling:
@mathewjordan Just made some updates which support `target.paintingFormat` as `string[]` values.   Thanks for the feedback.

wykhuh:
We are working on  a pdf viewer and newspaper viewer for Clover. We will make the plugins opensource  so anyone can add them to their Clover app.   Can you add an options key to component to make the  plugins customizable? Maybe something like:

```
<Viewer
  iiifContent="https://manifest-url-here"
  customDisplays={[
    {
      target: {
        canvasIds: ["http://a-canvas-id-in-manifest/access/0"],
      },
      component: { 
         name: MyDisplayComponent,
         options: { backgrondColor: 'red', zoomLevel: '5'}
      }
    },
  ]}
/>
```

mathewjordan:
> We are working on a pdf viewer and newspaper viewer for Clover. We will make the plugins opensource so anyone can add them to their Clover app. Can you add an options key to component to make the plugins customizable? Maybe something like:
> 
> ```
> <Viewer
>   iiifContent="https://manifest-url-here"
>   customDisplays={[
>     {
>       target: {
>         canvasIds: ["http://a-canvas-id-in-manifest/access/0"],
>       },
>       component: { 
>          name: MyDisplayComponent,
>          options: { backgrondColor: 'red', zoomLevel: '5'}
>       }
>     },
>   ]}
> />
> ```

Hi @wykhuh. Could these not be props on the custom component?

```jsx
<Viewer
  iiifContent="https://manifest-url-here"
  customDisplays={[
    {
      target: {
        paintingFormat: ["application/pdf"]
      },
      component: <MyCustomComponent options={{}} />,
    },
  ]}
/>
```

The alternative if we don't want to pass it as JSX, perhaps we just do a generic `props` property and then spread out whatever it contains?

```jsx
<Viewer
  iiifContent="https://manifest-url-here"
  customDisplays={[
    {
      target: {
        canvasIds: ["http://a-canvas-id-in-manifest/access/0"],
      },
      component: { 
         name: MyCustomComponent,
         props: {
           options: { 
             backgroundColor: 'red',
             zoomLevel: '5'
           }
         }
      }
    },
  ]}
/>
```

Rendering generically as...

```jsx
<MyCustomComponent {...props} />
```

adamjarling:
Hey @mathewjordan @wykhuh I just pushed up a commit which allows custom props to be passed through to a client component.  Something like this work?

https://github.com/samvera-labs/clover-iiif/pull/149/commits/663689cc37480627d5735d600dac230ce17081ad

adamjarling:
@mathewjordan @wykhuh Just added another commit which removes `painting` as a `prop` passed automatically into a client's Custom Display component.  Feels more clean to me to keep Clover only passing minimal data to avoid any slippery slopes, unless it's super convenient to pass something along and saves the client a headdache.   

So the Viewer now passes to a Custom Display component:

- The canvas `id`
- Any custom props the client Custom Display component wishes to pass through: `...restProps`
- ~`painting`~

https://github.com/samvera-labs/clover-iiif/pull/149/commits/db6914c276bfadb1ad47903d7251d9bafa0922fa

adamjarling:
If this is ready to merge in, I think the last commit @mathewjordan will be reverting the Viewer documentation to use the original `demo.mdx` code, and make a reference to how a user could get their hands on a customization example.  Maybe even a StackBlitz or CodeSandbox playground example referenced in the docs in addition to leaving this file in the repo: `docs/components/CustomDisplays/PDFViewer.tsx`

wykhuh:
@adamjarling If 

> @mathewjordan @wykhuh Just added another commit which removes `painting` as a `prop` passed automatically into a client's Custom Display component. Feels more clean to me to keep Clover only passing minimal data to avoid any slippery slopes, unless it's super convenient to pass something along and saves the client a headdache.
> 
> So the Viewer now passes to a Custom Display component:
> 
> * The canvas `id`
> * Any custom props the client Custom Display component wishes to pass through: `...restProps`
> * ~`painting`~
> 
> [db6914c](https://github.com/samvera-labs/clover-iiif/commit/db6914c276bfadb1ad47903d7251d9bafa0922fa)


1)  In `docs/components/CustomDisplays/PDFViewer.tsx`,  the iiifContent url is hard-coded, which  won't work for a production-ready component. I'm assuming we would use `componentProps` to pass in the  iiifContent url. 
2) If the  Custom Display component loads  iiifContent, then there will be multiple http get request to iiifContent url. (1st in `src/components/Viewer/index.tsx`, and 2nd in the Custom Display component). Is extra networks requests an issue?
3) `docs/components/CustomDisplays/PDFViewer.tsx`  is called 3 times, which means three http requests to iiifContent url. Not sure why that is happening.

adamjarling:
@wykhuh Thanks for the feedback.  I'll go through the comments below...
> 
> 1. In `docs/components/CustomDisplays/PDFViewer.tsx`,  the iiifContent url is hard-coded, which  won't work for a production-ready component. I'm assuming we would use `componentProps` to pass in the  iiifContent url.

Yea, via `display.componentProps` could be one way to pass the Manifest URI into a custom component.  I assumed that since this URI is the same value passed into the top level instance of `Viewer` through `iiifContent` prop (`<Viewer iiifContent="" />`, then it'd be accessible by the custom display component.  But I suppose if there are multiple IIIF Manifests your application is managing, it'd be easier to get direct access too.  

Perhaps this example CustomDisplay component is being biased by it's use of (or my misuse of 😀) the `vault` package too though.  This code assumes `vault` needs the root IIIF Manifest to begin the path towards parsing content, digging for `canvas` > `annotation` > `body`.  If your IIIF canvas URIs are dereferenceable, then you wouldn't need the manifest URI, and could just grab the Canvas itself via a network request (using `vault`, `fetch`, etc).

> 2. If the  Custom Display component loads  iiifContent, then there will be multiple http get request to iiifContent url. (1st in `src/components/Viewer/index.tsx`, and 2nd in the Custom Display component). Is extra networks requests an issue?

Mmmm, I suppose this is a valid point.  And would ultimately depend upon the size of the Manifest or Canvas being fetched?  We wouldn't consider it a big performance hit, but if this is on your end, then I'm open to passing some data straight through to a custom display component (like `painting`, or even the `canvas` itself if that helps?

> 3. `docs/components/CustomDisplays/PDFViewer.tsx`  is called 3 times, which means three http requests to iiifContent url. Not sure why that is happening.

This was just due to a quick, lazy implementation on my end.  I pushed a commit, wrapping the network request in a `useEffect` hook to avoid unnecessary network requests.

> 4. In `src/components/Viewer/Painting/Painting.tsx`,  if `painting.length >0`, Clover will show a select menu that lets people choose which painting to show. If I'm reading the code correctly, that select menu won't work with Custom Display since Custom Display doesn't have access to `annotationIndex` or `painting`.

I'll look into this...

adamjarling:
@mathewjordan @wykhuh 

> 4. In `src/components/Viewer/Painting/Painting.tsx`,  if `painting.length >0`, Clover will show a select menu that lets people choose which painting to show. If I'm reading the code correctly, that select menu won't work with Custom Display since Custom Display doesn't have access to `annotationIndex` or `painting`.

I think this behavior may have some issues independent of this Custom Display work?   Looks like the Select Choice was intended to support https://github.com/samvera-labs/clover-iiif/blob/main/public/manifest/cookbook/0033-choice.json. Do you think you'd use this feature @wykhuh ?   The way I read this code, a custom display doesn't need to know about `annotationIndex` or `painting`, as that's handled by `Painting` in this new block:  https://github.com/samvera-labs/clover-iiif/blob/f91d4efd06cc788b27f4d7cadca12d97e18d11a0/src/components/Viewer/Painting/Painting.tsx#L56. So I *think* this code would hold up, but maybe it could be more defensive in displaying options without labels, dark mode/light mode and making sure it flows nicely in the layout w/ a Custom Display component.

I'd suggest opening a separate ticket to address these concerns if ya'll are ok with it?

wykhuh:
@adamjarling @mathewjordan 

> @wykhuh Thanks for the feedback. I'll go through the comments below...
> 
> > 1. In `docs/components/CustomDisplays/PDFViewer.tsx`,  the iiifContent url is hard-coded, which  won't work for a production-ready component. I'm assuming we would use `componentProps` to pass in the  iiifContent url.
> 
> Yea, via `display.componentProps` could be one way to pass the Manifest URI into a custom component. I assumed that since this URI is the same value passed into the top level instance of `Viewer` through `iiifContent` prop (`<Viewer iiifContent="" />`, then it'd be accessible by the custom display component (which is why it's hardcoded in the example). But I suppose if there are multiple IIIF Manifests your application is managing, it'd be easier to get direct access to the specific resource.
> 
> But, perhaps this example CustomDisplay component is being biased by it's use of (or my misuse of 😀) the `vault` package. This code assumes `vault` needs the root IIIF Manifest to begin the path towards parsing content, digging for `canvas` > `annotation` > `body`. If your IIIF canvas URIs are dereferenceable, then one may not need the manifest URI, and could just grab the Canvas itself via the canvas `id` value being passed back to the custom component via a network request (using `vault`, `fetch`, etc).
> 

Currently, we can't grab the Canvas itself using the  canvas id. So we would need pass in Manifest URI as componentProps,  to do a network request for the Manifest URI, and parse the returned data  in order to get the width and height for the media.

> > 2. If the  Custom Display component loads  iiifContent, then there will be multiple http get request to iiifContent url. (1st in `src/components/Viewer/index.tsx`, and 2nd in the Custom Display component). Is extra networks requests an issue?
> 
> Mmmm, I suppose this is a valid point. And would ultimately depend upon the size of the Manifest or Canvas being fetched? We wouldn't consider it a big performance hit, but if this is on your end, then I'm open to passing some data straight through to a custom display component (like `painting`, or even the `canvas` itself if that helps?
> 

For the newspaper viewer,  we have multiple page newspaper issues with OCR text for each page. We are considering displaying all OCR text in Clover for each page by using annotations. If we include the OCR text in root manifest, the manifest will be a decent size.

adamjarling:
@wykhuh Gotcha.  Was chatting with @mathewjordan this morning, and it does sound like the case is building to feed the `painting` (now called `annotationBody`), to a Custom Display component, alongside the Canvas `id` and pass-through, user-defined `props`.   Just pushed up a commit, and updated the `PDFViewer` example component, which does simplify it quite a bit.   Glad we chatted through this. 👍

wykhuh:
> @mathewjordan @wykhuh
> 
> > 4. In `src/components/Viewer/Painting/Painting.tsx`,  if `painting.length >0`, Clover will show a select menu that lets people choose which painting to show. If I'm reading the code correctly, that select menu won't work with Custom Display since Custom Display doesn't have access to `annotationIndex` or `painting`.
> 
> I think this behavior may have some issues independent of this Custom Display work? Looks like the Select Choice was intended to support https://github.com/samvera-labs/clover-iiif/blob/main/public/manifest/cookbook/0033-choice.json. Do you think you'd use this feature @wykhuh ?
> 
> The way I read this code, a custom display doesn't need to know about `annotationIndex` or `painting`, as the targeting is handled by `Painting` here:
> 
> https://github.com/samvera-labs/clover-iiif/blob/e6cace2ab8b8ac23a2ec52334872e51c6b88ee34/src/components/Viewer/Painting/Painting.tsx#L57
> 
> I _think_ this code would hold up, but maybe it could be more defensive in displaying options without labels, dark mode/light mode and making sure it flows nicely in the layout w/ a Custom Display component.
> 
> I'd suggest opening a separate ticket to address these concerns if ya'll are ok with it?

@adamjarling For my current projects, I we don't have a need for Select Choice. I only noticed it because I was browsing the Painting  component and noticed the select menu.  If you have projects that need an improved Select Choice, then creating a separate ticket makes sense.

pages/docs/viewer.mdx
```diff
@@ -333,3 +347,81 @@ return (
...
+    }]
+  />
+```
+
+The `Viewer` component will pass the following props to your custom display component:
+
+- `id`: The canvas `id` for the resource being rendered. This may be helpful if you wanted to use the canvas `id` to fetch additional data from your application's API.
+- `annotationBody`: The `body` value for a canvas `Annotation` item with `motivation` "painting".
+
+```json {5-11}
```
mathewjordan, line 413:
```suggestion
```json {5-10}
```

Just a line off here.

mathewjordan (approved):
Looking good to me still. One thought I have for the future... We may want to explore how to make the demo page conducive to representing our various props.

PR #152, opened 2023-12-04
Update CORS setting on Thumbnail component

adamjarling:
## What does this do?
Add `crossorigin` attribute to underlying `img` tag in Thumbnail, to help with CORS display errors.  The problem is that Canopy doesn't load `img` resources in StackBlitz sandbox environments.

See this issue for more context:
https://github.com/nulib/dc-api-v2/issues/178

## How to test?
See this StackBlitz instance (open in Firefox and Safari browsers to see the potential fix):
https://stackblitz.com/edit/cors-corp-image-example-pzxncq

### DIFF
--- a/src/components/Primitives/ContentResource/ContentResource.tsx
+++ b/src/components/Primitives/ContentResource/ContentResource.tsx
@@ -116,6 +116,7 @@ const ContentResource: React.FC<PrimitivesContentResource> = (props) => {
         <StyledResource
           as="img"
           alt={alt}
+          crossOrigin="anonymous"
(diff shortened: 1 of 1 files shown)


mathewjordan:
Hey, @adamjarling, do we still want this?

adamjarling:
> Hey, @adamjarling, do we still want this?

@mathewjordan Let's close this with no action if easier.  If there pops up a use case to demo an app which consumes Clover in the future, in Stackblitz, or maybe CodePen or something, then maybe we re-address?

PR #153, opened 2023-12-07
Bug fix to play audio files

adamjarling:
## What does this do?

- Fixes a bug which was overlaying the "audio visualizer" element on top of the `video` element when playing audio files, thereby not allowing a user to play audio files in the UI.
- Fix an audio visualizer component bug when a user clicks around to different time values in the `video` element

image

## How to test
- Spin up the app in local dev environment
- Test this link and verify you can play the audio file in the UI, and the visualizer appears:

http://localhost:3000/docs/viewer/demo?iiif-content=https://iiif.io/api/cookbook/recipe/0002-mvm-audio/manifest.json

### DIFF
--- a/src/components/Viewer/Player/AudioVisualizer.tsx
+++ b/src/components/Viewer/Player/AudioVisualizer.tsx
@@ -1,76 +1,85 @@
 /* eslint-disable react/display-name */
 
+import React, { useCallback } from "react";
+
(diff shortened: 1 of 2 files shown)


markpbaggett (approved):
@adamjarling This looks great to me.  

I actually noticed this the other day with one of my audio works, but assumed I had broken something in my manifest.  Didn't even think to look at zindex.

PR #158, opened 2023-12-12
Address new OSD instantiation drift.

mathewjordan:
The places the instance instantiation within a useEffect to ensure new instances are only created if the the URI of the image being loaded into OpenSeadragon changes.

Previous to this, any higher level re-renders of Clover were forcing a new UUID to be generated.


adamjarling (approved):
Tested the release candidate and works great.  👍

PR #161, opened 2023-12-19
Add CSS helper classes

adamjarling:
## What does this do?
Adds CSS helper classes which consuming applications can target to further customize the Clover Viewer.

## How to test
- General opinion on the class naming convention
- Run the test suite (tests added to verify class names exist and don't get knocked out of place)

## General notes
@mathewjordan This might seem like a small change, but Clover commits to providing this vehicle for customization, its pretty important to clearly announce/document any changes to class names so as not to break client app styling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # dev
 .build
 .next
+.vscode/
 dist
 node_modules
 out
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,0 @@
-{
-  "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
-  },
-  "editor.formatOnSave": true,
-  "editor.defaultFormatter": "esbenp.prettier-vscode"
-}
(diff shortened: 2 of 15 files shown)


mathewjordan (approved):
I like the additions. Should we document helper classes (briefly) in the Viewer docs? Maybe not call out each distinct class name but state something generic regarding their inclusion?

adamjarling:
> I like the additions. Should we document helper classes (briefly) in the Viewer docs? Maybe not call out each distinct class name but state something generic regarding their inclusion?

@mathewjordan Yes, great call.  I'll push up an update.

adamjarling:
How does this look @mathewjordan ? https://github.com/samvera-labs/clover-iiif/pull/161/commits/97fc034afaa920e8cebd4dfd5e4b8278984c67d3

mathewjordan:
> How does this look @mathewjordan ? [97fc034](https://github.com/samvera-labs/clover-iiif/commit/97fc034afaa920e8cebd4dfd5e4b8278984c67d3)

Solid to me. 🚀

PR #163, opened 2024-01-03
Update dependencies

adamjarling:
## What does this do?

There were some major dependency version lags, so this PR attempted to upgrade all deps.  However, after upgrading to Vite `v5`, the TypeScript types generation as part of the build process broke.  We're using `dts-bundle-generator` to handle this.  I tried updating the build to use `vite-plugin-dts`, a more popular solution in numbers, but that was erroring-out under the hood as well when trying to roll up all types into one file. 

So, the end solution here is to pin to an older version of `dts-bundle-generator`.  

Another byproduct of upgrading to the latest Vite, is that the web-component build is broken as well.  I'll look at this and see if there's a way to compile it, and make sure it doesn't break for anyone using it.   Is anyone using it I wonder tho?

## How to test
Pull in the latest release candidate version into a demo environment:
```bash
pnpm add @samvera/clover-iiif@next
```
And verify
- The Clover components work as expected
- Typescript intellisense is working in your IDE
- You can use the web-component version as well


mathewjordan:
Awesome to get things more up to date. I'm not sure about any usage as far as the web components. We never advertised this beyond the section in the docs. The one benefit I see to continue packaging web components is that we may have future use for it within notebooks (jupyter or livebook) as we trend towards that style of development and storytelling.

adamjarling:
Web component bundling now working @mathewjordan .  There is some TypeScript technical debt needed to address in the future, and I'll create a new issue for it internally in our Northwestern board.

- Vite 5 uses a new Rollup version, which has deprecated non ESModule imports (i.e `require (...)`).  So we need to update `compilerOptions.moduleResolution` to be `Bundler` or `Node16` in `tsconfig.json`.   Doing so now breaks the build, mostly a ton of Stitches components which need explicit type annotations, so will leave this for future work.
- Update `dts-bundle-generator` to the latest package (it's pinned to one major version behind, which is the only one which will still work for some reason).  It might have something to do with the above bullet point.

mathewjordan (approved):
🚀

PR #166, opened 2024-01-16
Feature/newspaper viewer  revisions (WIP)

mathewjordan:
Just a work in progress...


pages/newspaper.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
```
wykhuh, line 1: There is a weird bug. When there is a collection, Clover will display a dropdown menu to switch between items in the collection. If you use the menu to switch items the first time, the canvas is updated. If you use the menu to switch items 2nd, 3rd, etc times, the manifest changes but the canvas painting does not change. 

http://localhost:3000/newspaper    

https://samvera-labs.github.io/clover-iiif/docs/viewer/demo?iiif-content=https://iiif.io/api/cookbook/recipe/0068-newspaper/newspaper_title-collection.json

src/components/Viewer/InformationPanel/Annotation/Item.tsx
```diff
@@ -0,0 +1,112 @@
...
+          />
+        );
+      case "text/vtt":
+        return (
+          <AnnotationItemVTT
+            label={annotationBody[0].label}
+            vttUri={annotationBody[0].id || ""}
+          />
+        );
+      default:
```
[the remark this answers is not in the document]
adamjarling: @wykhuh Glad you mentioned this, as the `renderItemBody` function could be cleaned up a bit.  I'm not the biggest fan of using a `switch` statements, but seems a decent enough use case for it here.  Since the `switch` is on a type `string`, your regex matcher above could return `undefined`, so not entirely sure it's valid, but TypeScript is not complaining and it does logically read better not to have the image regex check tucked under the `default` (this seems wrong too).  I'll plug your code in 😀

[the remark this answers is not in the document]
adamjarling: That being said, I'm going to slightly re-factor this component and simplify the `renderItemBody` function to only do one thing (`switch`).

pages/newspaper.tsx
adamjarling: @wykhuh @mathewjordan Spent the morning digging into this odd behavior, and getting quite familiar with `vault`s storage structure.  Which finally pointed me to the underlying problem, or at least the cause of this bug.  `vault` will store it's resources, ("Collection", "Annotation", "AnnotationPage", etc) keyed by the resource `id`.  There were duplicate resource `id`s (URIs), for an `Annotation` and `AnnotationPage` between the two fixture files: 
- public/manifest/newspaper/newspaper_issue_1.json
- public/manifest/newspaper/newspaper_issue_2.json

When switching to the 2nd manifest in `public/manifest/newspaper/newspaper_collection.json`, `vault` was overriding the value for:
- `http://localhost:3000/manifest/newspaper/annotation_page_painting/ap1`
- `http://localhost:3000/manifest/newspaper/annotation/p1`

In commit d73bbeae21187ead8beeee25e836575ef222cfb3, updating the `id`s in `newspaper_issue_2.json` makes `vault` happy and it can uniquely retrieve items as expected once again.

wykhuh: @adamjarling The bug also happens in the IIIF newspaper recipe  https://iiif.io/api/cookbook/recipe/0068-newspaper/newspaper_title-collection.json   Is the bug also caused by a duplicate id in the IIIF  recipe?

adamjarling: @wykhuh Hmmm, well that is interesting.  I don't see any duplicate `id`s in that manifest, so that's not the problem, but something else is.  I think we should add that recipe to the `annotations` test page and look into further.

adamjarling:
Merged PR of this work:  https://github.com/samvera-labs/clover-iiif/pull/170

PR #170, opened 2024-02-15
Annotations support

adamjarling:
## What does this do?

Improved support for Annotations in the `InformationPanel` component.  

image

See this example IIIF Manifest to view initial support for IIIF Annotations with `motivation` values beyond "supplementing".

`public/manifest/newspaper/newspaper_issue_1.json`

The interactive "annotation" component rendered in the UI will be determined by the `body.format` value of the `annotation` item.  For example, if a Manifest's `canvas.annotations[0].items[0]` looks like:

```json
         {
              "id": "http://localhost:3000/manifest/newspaper/newspaper_issue_1-anno_p1.json-1",
              "type": "Annotation",
              "motivation": "highlighting",
              "body": {
                "type": "TextualBody",
                "format": "text/plain",
                "language": "de",
                "value": "Berliner"
              },
              "target": "http://localhost:3000/manifest/newspaper/canvas/i1p1#xywh=839,3259,118,27"
            }
```

...the UI will cue off of  it's`body.format` value, or "text/plain".   

In this PR, the UI supports values: 
- `text/plain`
- `text/html`
- `text/vtt`
- any type of image, `image/*`


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-pnpm exec lint-staged
+# pnpm exec lint-staged
```
wykhuh, line 4: Should this comment be removed before the PR is merged?

adamjarling: @wykhuh Yes, good catch.  Thanks.

adamjarling:
Also, once merged I can package up a release candidate, and we can sanity check some installations?

wykhuh:
> Also, once merged I can package up a release candidate, and we can sanity check some installations?

@adamjarling Did you figure out what was causing the bug of canvas not changing when people use the drop down menu to change items in a collection? That bug can be dealt with in a separate PR.

adamjarling:
> > Also, once merged I can package up a release candidate, and we can sanity check some installations?
> 
> @adamjarling Did you figure out what was causing the bug of canvas not changing when people use the drop down menu to change items in a collection? That bug can be dealt with in a separate PR.

@wykhuh Yea, I think the root cause of the behavior is still duplicate `canvas` ids.  The canvas `id` value, "https://iiif.io/api/cookbook/recipe/0068-newspaper/canvas/p1", is duplicated across the 2 manifests in the Collection for [Recipe 0068](https://iiif.io/api/cookbook/recipe/0068-newspaper/):

- https://iiif.io/api/cookbook/recipe/0068-newspaper/newspaper_issue_1-manifest.json
- https://iiif.io/api/cookbook/recipe/0068-newspaper/newspaper_issue_2-manifest.json

Because we're parsing IIIF resources with `vault` library, vault tracks it's state with `id` values as keys and the parsing becomes inaccurate in certain situations with duped keys/ids.

In a production environment, seems these values should be unique, and dereferencable.   Chatted briefly with @mathewjordan on it, and yea if this is a real issue, let's create a new PR for it.

src/components/Viewer/ImageViewer/OSD.tsx
```diff
@@ -62,16 +90,32 @@ const OSD: React.FC<OSDProps> = ({ uri, hasPlaceholder, imageType }) => {
...
-          );
+            });
+            dispatch({
+              type: "updateOpenSeadragonViewer",
+              openSeadragonViewer: viewer,
+            });
+            if (configOptions.annotationOverlays?.renderOverlays) {
+              addOverlaysToViewer(viewer, canvas, configOptions, annotations);
+            }
+          });
```
mathewjordan, line 118: We may want to abstract this out at some point.

src/lib/openseadragon-helpers.ts
```diff
@@ -0,0 +1,239 @@
...
+    const overlay = OsdSvgOverlay(viewer);
+    overlay.node().append(newElement);
+    overlay._svg?.setAttribute("class", "annotation-overlay");
+
+    // BUG: svg with children elements aren't formated correctly.
+    child.childNodes.forEach((child) => {
+      svg_processChild(viewer, child, configOptions, scale);
+    });
+  }
+}
```
mathewjordan, line 165: Should we have an `svg-helpers` file for all of these prefixed functions? I understand that they likely will only be used for openseadragon, but just curious here.

mathewjordan (approved):
Let's roll the RC. 🏎️

src/components/Viewer/InformationPanel/Annotation/Image.tsx
```diff
@@ -0,0 +1,23 @@
...
+};
+
+const AnnotationItemImage: React.FC<AnnotationItemImageProps> = ({
+  caption,
+  handleClick,
+  imageUri,
+}) => {
+  return (
+    <ButtonStyled onClick={handleClick}>
+      <img src={imageUri} alt={`A visual annotation for ${caption}`} />
```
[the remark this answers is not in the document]
wykhuh: Clover has a few places with hard coded text.  Some of CollectiveAccess users need the site to be in languages other than English.  Will Clover eventually add support for localization?

wykhuh:
> @wykhuh Yea, I think the root cause of the behavior is still duplicate `canvas` ids. The canvas `id` value, `https://iiif.io/api/cookbook/recipe/0068-newspaper/canvas/p1`, is duplicated across the 2 manifests in the Collection for [Recipe 0068](https://iiif.io/api/cookbook/recipe/0068-newspaper/):
> 
> * https://iiif.io/api/cookbook/recipe/0068-newspaper/newspaper_issue_1-manifest.json
> * https://iiif.io/api/cookbook/recipe/0068-newspaper/newspaper_issue_2-manifest.json
> 
> Because we're parsing IIIF resources with the `vault` library, vault tracks it's state with `id` values as keys. The parsing becomes inaccurate in certain situations with duped keys/ids.
> 
> In a production environment, seems these values should be unique, and dereferencable. Chatted briefly with @mathewjordan on it, and yea if this is a real issue, let's create a new PR for it.

@adamjarling  I checked Mirador, Universal Viewer and Annona. They can switch between newspaper issues despite the non-unique ids. I don't know if this is a case of the other viewers being too lax or the `vault` being too strict about unique ids.   I don't think CollectiveAccess needs to use the drop down menu in Clover to switch between issues, so this bug isn't a high priority for us. I'll let y'all decide to either leave things as is or deal with the issue.

adamjarling:
> > @wykhuh Yea, I think the root cause of the behavior is still duplicate `canvas` ids. The canvas `id` value, `https://iiif.io/api/cookbook/recipe/0068-newspaper/canvas/p1`, is duplicated across the 2 manifests in the Collection for [Recipe 0068](https://iiif.io/api/cookbook/recipe/0068-newspaper/):
> > 
> > * https://iiif.io/api/cookbook/recipe/0068-newspaper/newspaper_issue_1-manifest.json
> > * https://iiif.io/api/cookbook/recipe/0068-newspaper/newspaper_issue_2-manifest.json
> > 
> > Because we're parsing IIIF resources with the `vault` library, vault tracks it's state with `id` values as keys. The parsing becomes inaccurate in certain situations with duped keys/ids.
> > In a production environment, seems these values should be unique, and dereferencable. Chatted briefly with @mathewjordan on it, and yea if this is a real issue, let's create a new PR for it.
> 
> @adamjarling I checked Mirador, Universal Viewer and Annona. They can switch between newspaper issues despite the non-unique ids. I don't know if this is a case of the other viewers being too lax or the `vault` being too strict about unique ids. I don't think CollectiveAccess needs to use the drop down menu in Clover to switch between issues, so this bug isn't a high priority for us. I'll let y'all decide to either leave things as is or deal with the issue.

@wykhuh Noted, and thanks for the clarifying.  For future reference, one eventual solution/fix might be to provide more unique `key` values on `Viewer` components, used to trigger re-renders.  Maybe something like:

```jsx
<SomeComponentNotRerendering key={MANIFEST_ID_AND_CANVAS_ID} />
```

PR #177, opened 2024-03-07
Add `informationPanel.vtt.autoScroll` option & functionality

mbklein:
### Changes

- Implement auto-scrolling for VTT captions
- Add `autoScroll` option to `informationPanel` config options
- Add documentation for new `autoScroll` option

### Notes

- The autoscroll functionality is hooked into the viewer context to allow for live toggling on and off. I tested this functionality in a browser with a checkbox wired up to dispatch the toggle event, but I couldn't figure out how to position/style it so it wasn't ugly, so there's currently no UI for the toggle.
- Autoscroll will frustrate any efforts to scroll manually through the captions as it will continue to scroll back to what's currently playing. If there's a way to detect a manual scroll action and toggle autoscroll off automatically, that would be ideal.
- The documentation could maybe be clearer.

I'm open to feedback (and pointers/assistance on those two UI bullet points if you're amenable).


src/context/viewer-context.tsx
```diff
@@ -21,6 +31,7 @@ export type ViewerConfigOptions = {
   canvasHeight?: string;
   ignoreCaptionLabels?: string[];
   informationPanel?: {
+    autoScroll?: AutoScrollSettings | boolean;
```
mathewjordan, line 34: Should we be more specific about what is auto scrolling? Something like `vttAutoScroll` just to keep things organization a little better?

or even...

```js
vtt?: {
  autoScroll?: {
    ...
  }
}
```

mathewjordan (changes requested):
@mbklein This is pretty solid and I think is good. I just have a few thoughts and concerns. 

In addition, I was toying around with adding an event handler to control `isUserScrolling`, allowing us an additional conditional:

```
  useEffect(() => {
    if (autoScroll?.enabled && isActive && ref.current && !isUserScrolling) {
      const scrollableParent = findScrollableParent(ref.current);
      if (scrollableParent && scrollableParent instanceof HTMLElement) {
        scrollableParent.scrollTo({
          top: ref.current.offsetTop,
          ...autoScroll?.settings,
        });
      }
    }
  }, [autoScroll, isActive]);
```

I'm open to that being an an iteration though. What are your thoughts?

mbklein:
I really like the idea of `isUserScrolling` - I had a similar idea, that if the user is scrolling, don't shift, but once they stop, go back to normal behavior. But I'd put the property on a delay; e.g. it would be true as soon as they start scrolling until about 1-1.5 seconds after they stop. That would give them time to start again if they weren't finished, or click a new caption to start playback from there. It would be really frustrating to arrive at the line you want only to have autoscroll jump back before you have a chance to click it.

src/context/viewer-context.tsx
mbklein: That's funny, I originally called it `autoScrollAnnotation` and then  `vttAutoScroll` before settling on `autoscroll`. I'm happy to change it or nest it; I was just trying to stay as simple and plain-English as possible.

mbklein:
I really like the idea of `isUserScrolling` - I had a similar idea, that if the user is scrolling, don't shift, but once they stop, go back to normal behavior. But I'd put the property on a delay; e.g. it would be true as soon as they start scrolling until about 1-1.5 seconds after they stop. That would give them time to start again if they weren't finished, or click a new caption to start playback from there. It would be really frustrating to arrive at the line you want only to have autoscroll jump back before you have a chance to click it.

mbklein:
> I wonder if isUserScrolling should be set at a higher level in the component where the Scrolling container is actually rendered, rather than on ever cue.

Great minds think alike, I guess. 😄 I was doing just that. I pushed a new commit that:

- Moves the "Is the user scrolling?” decision up to the `Scroll` component
- Adds `viewerState` awareness to `Cue` to avoid having to pass the auto scroll config down through the tree, and also so it can update on when it's autoscrolling (because `onScroll` can't distinguish between a user-generated scroll event and one that came from a `scrollTo()` or `scrollIntoView()` call)
- Moves the `autoScroll` option under `informationPanel.vtt`

I fear this solution might be abusing the viewer state a little bit, but that seemed to be the most effective way to get the Scroll and Cue components to communicate their scrolling-related info to each other. Working great in practice.

src/context/viewer-context.tsx
```diff
@@ -49,6 +67,7 @@ const defaultConfigOptions = {
   canvasHeight: "61.8vh",
   ignoreCaptionLabels: [],
   informationPanel: {
+    autoScroll: false,
```
mathewjordan, line 70: Following the new config shape, this should be adjusted to something like:

Note that I set enabled to `true`, I think that it'd be nice for that to be default.

```suggestion
    vtt: {
      autoScroll: {
        enabled: true,
        settings: defaultAutoScrollSettings,
      },
    },
```

src/components/Viewer/InformationPanel/Annotation/VTT/Cue.tsx
```diff
@@ -1,16 +1,45 @@
...
+
+    element = element.parentNode as HTMLElement;
+  }
+
+  return null;
+};
+
 const Cue: React.FC<Props> = ({ label, start, end }) => {
+  const dispatch: any = useViewerDispatch();
+  const { autoScroll, userScrolling }: ViewerContextStore = useViewerState();
```
mathewjordan, line 40: I'm not sure exactly what's going on, it looks like this isn't peeling off the autoScroll correctly, at least in my local. For it to work for me, I need to dig into `configOptions.informationPanel.vtt.autoScroll`.

src/context/viewer-context.tsx
```diff
@@ -73,25 +92,36 @@ export type CustomDisplay = {
...
 
+export interface UserScrollingState {
+  state: boolean;
+  timeout: number | undefined;
+}
+
 export interface ViewerContextStore {
   activeCanvas: string;
   activeManifest: string;
+  autoScroll?: AutoScrollOptions;
```
mathewjordan, line 103: Should this be here? Looks like it is duplicating the configOption.

src/context/viewer-context.tsx
```diff
@@ -73,25 +92,36 @@ export type CustomDisplay = {
...
   informationOpen: boolean;
   isLoaded: boolean;
+  userScrolling?: UserScrollingState;
   vault: Vault;
   openSeadragonViewer: OpenSeadragon.Viewer | null;
 }
 
 export interface ViewerAction {
   type: string;
+  autoScroll: boolean;
```
mathewjordan, line 117: Same, dup.

src/context/viewer-context.tsx
```diff
@@ -100,15 +130,31 @@ export interface ViewerAction {
 export const defaultState: ViewerContextStore = {
   activeCanvas: "",
   activeManifest: "",
+  autoScroll: expandAutoScrollOptions(
+    defaultConfigOptions?.informationPanel?.autoScroll,
+  ),
```
mathewjordan, line 135: Same, dup. Or perhaps I'm not fully understanding the purpose of putting this here as well as on the `configOption.informationPanel.vtt.autoScroll`.

mathewjordan (changes requested):
Hi @mbklein. Quicky review again before the day really starts. This looks good other than my concerns about duplicating the autoScroll config in the viewer-context.

src/context/viewer-context.tsx
mbklein: I made some assumptions regarding the use of `configOptions`, the rest of the `ViewerContextStore`, and `ViewerAction`:

- `ViewerContextStore.configOptions` – the config that came in from the `ViewerProvider`. I treat it as read-only in case we need a “reset to initial state” trigger somewhere
- `ViewerContextStore.autoScroll` – the _current_ autoScroll behavior, which may have been changed programatically (I took this "original state/current state" pattern from the duplication of `configOptions.informationPanel.open` and `informationOpen`)
- `ViewerAction.autoScroll` – the new autoscroll setting coming in as part of a dispatch to update `ViewerContextStore.autoScroll.enabled`. Maybe it should be a full `AutoScrollOptions` object instead of just the boolean so it can all be updated programmatically instead of just the `enabled` part.

If it's OK for the current state to just overwrite the original state, I have no problem using `configOptions.informationPanel.autoScroll` throughout.

Other options:
- `autoScrolling` is the indicator of whether a `Cue` is currently in the middle of scrolling itself into view. It needed to be a top-level bool within the state because nesting it with other `autoScroll` stuff would trigger too many dependencies when it changes.
- `userScrolling` stores whether there's a manual user scroll happening, as well as the reference ID of the `setTimeout` that will reset that value. We could probably just use the presence of the timeout's ID to determine truthiness (e.g., `userScrolling === undefined` replaces `!userScrolling.enabled`), but this was more explicit.

src/context/viewer-context.tsx
mbklein: Another reason for separating the `configOptions` scroll setting from the top-level scroll setting in the state is to allow for `true` and `false` config values that automatically expand themselves to `{ enabled: [true|false], settings: DefaultSettings }`.

src/context/viewer-context.tsx
mathewjordan: You're correct in that `configOptions` is read only.

I guess I was misunderstanding why `autoScroll` was being set at the top of the context as well. I could see a scenario where a user does toggle this on/off. But in this case, I think it would just be a boolean and not the whole autoScroll object. I hope I'm getting that correct now.

Some of my confusion here is self inflicted as a lot of the naming conventions here at the top of the config need to be thought out a bit better. This isn't on you but we could set a new standard here and now for this type of state variable.

I think we could go forward with more descriptive props like:

 - `isAutoScrollEnabled`
 - `isAutoScrolling`
 - `isUserScrolling`
 
 AND
 
 - `isInformationOpen`

mathewjordan:
Hey @mbklein 

See my comment above for https://github.com/samvera-labs/clover-iiif/pull/177#discussion_r1522405234. I'm happy to chat through it tomorrow if we can have a chance to do so. I think we're close.

src/context/viewer-context.tsx
mbklein: OK, this makes sense to me now.

src/components/Viewer/InformationPanel/InformationPanel.styled.tsx
```diff
@@ -83,7 +84,23 @@ const Content = styled(Tabs.Content, {
...
+  handleScroll,
+  children,
+  className,
+}) => (
+  <div className={className} onScroll={handleScroll}>
+    {children}
+  </div>
+);
+
+const Scroll = styled(ScrollableComponent, {
```
mathewjordan, line 103: 👍

src/components/Viewer/InformationPanel/InformationPanel.tsx
```diff
@@ -6,14 +6,20 @@ import {
...
+  useViewerState,
+} from "src/context/viewer-context";
 
 import AnnotationPage from "src/components/Viewer/InformationPanel/Annotation/Page";
 import { AnnotationResources } from "src/types/annotations";
 import Information from "src/components/Viewer/InformationPanel/About/About";
 import { InternationalString } from "@iiif/presentation-3";
 import { Label } from "src/components/Primitives";
 
+const UserScrollTimeout = 1500; // 1500ms without a user-generated scroll event reverts to auto-scrolling
```
mathewjordan, line 21: We can iterate on this further but this value would be nice to add to the configOptions.

src/components/Viewer/Viewer/Content.tsx
```diff
@@ -30,7 +30,7 @@ const ViewerContent: React.FC<ViewerContentProps> = ({
   items,
   painting,
 }) => {
-  const { informationOpen, configOptions } = useViewerState();
+  const { isInformationOpen, configOptions } = useViewerState();
```
mathewjordan, line 33: Thanks for updating this.

mathewjordan (approved):
Looking really good. This is a great addition. :shipit:

PR #178, opened 2024-03-11
Always render annotation resources in Viewer

adamjarling:
## What does this do?

Fixes a bug where Annotation Resources were being hidden when a user configures their `Viewer` to hide the `InformationPanel`.  For example, say there are VTT files associated with a Video in a `canvas`, these should always render alongside the content.

image

## How to test:
Configure the viewer to hide the `InformationPanel`, i.e.
```jsx
<Viewer
  options={{
    canvasHeight: "640px",
    informationPanel: {
      open: false,
      renderAbout: false,
      renderToggle: false,
    },
  }}
/>
```

If running Clover in a dev environment, use this manifest:

https://api.dc.library.northwestern.edu/api/v2/works/44f70212-029f-44cd-a3b3-03c18b208103?as=iiif

...or any manifest which contains Annotation Resources (doesn't have to be `VTT` annotations).

### DIFF
--- a/src/components/Viewer/Viewer/Content.tsx
+++ b/src/components/Viewer/Viewer/Content.tsx
@@ -15,7 +15,7 @@ import Painting from "../Painting/Painting";
 import React from "react";
 import { useViewerState } from "src/context/viewer-context";
 
-interface ViewerContentProps {
+export interface ViewerContentProps {
   activeCanvas: string;
   annotationResources: AnnotationResources;
   isAudioVideo: boolean;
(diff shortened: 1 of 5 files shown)


src/components/Viewer/Viewer/Content.tsx
```diff
@@ -67,7 +67,7 @@ const ViewerContent: React.FC<ViewerContentProps> = ({
           </MediaWrapper>
         )}
       </Main>
-      {informationOpen && isAside && (
+      {(informationOpen || isAside) && (
```
mathewjordan, line 70: 👍

mathewjordan (approved):

Issue #179, opened 2024-03-11
With one video in an iFrame, Clover's canvas doesn't fill the container.


tpendragon:
We embed viewers via iFrames, so we rely on viewers filling the container they're in. It looks like clover doesn't have a containing div that expands to fill, and instead you can pass the height of the canvas (which may vary.) I've chatted about this a little with y'all, but thought it'd be good to record a ticket.

![Screen Shot 2024-03-11 at 12 13 35 PM](https://github.com/samvera-labs/clover-iiif/assets/2806645/5707c60e-94cf-4c2d-a9b1-1979b4028ed4)

PR #185, opened 2024-03-13
Scale viewer to its container.

tpendragon:
Work towards #179 

I'm very new to this project, so I don't know what kind of tests y'all do, but this is my attempt at getting the CSS to adjust to its parent container. It looks like it works to me, if I shove the demo page into a tiny container.

### DIFF
--- a/src/components/Viewer/ImageViewer/ImageViewer.styled.tsx
+++ b/src/components/Viewer/ImageViewer/ImageViewer.styled.tsx
@@ -5,7 +5,7 @@ const Navigator = styled("div", {
   zIndex: "1",
   top: "1rem",
   left: "1rem",
-  width: "161.8px",
+  maxWidth: "161.8px",
   height: "100px",
   backgroundColor: "#000D",
   boxShadow: "5px 5px 5px #0002",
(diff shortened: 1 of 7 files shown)


tpendragon:
@eliotjordan did some testing for our use case and says this didn't make it any better, so I'm closing.

mathewjordan:
@tpendragon Can we chat to talk it through? The goal is to have Clover extend to the width of its wrapping container on the client side, correct? Similar to how Mirador behaves?

So, in this example, Clover's canvas would extend it's height and width automatically to fit. The thumbnails for canvases would take a certain amount of that height, but the canvas height would grow accordingly (using flexbox).

```jsx
<div style="height: 1000px; width: 100%">
  <CloverIIIFViewer />
</div>
```

mathewjordan:
Sweet, so here is the issue I authored... My approach was going to be this....

---

## Description

There is a use case where Clover needs to expand to the height of its wrapping container rather than have it be set by a fixed height in the config. As a side note, auto height expansion is how Mirador works.


Update type of canvasHeight to be more specific.
https://developer.mozilla.org/en-US/docs/Web/CSS/height

```ts
canvasHeight?: React.CSSProperties["height"];
```

If `canvasHeight === "auto", then we set a class at a high level within Clover and use flexbox to grow the canvasHeight to its bounds. We then also need to set  Clover to absolutely position itself and expand to the bounds of the consuming apps wrapping container.

## Done Looks Like

- [ ] Clover has auto height expansion from the settable config
- [ ] Clover can also be set by a fixed height (backwards compatibility for current implementations)

tpendragon:
@mathewjordan Looks great!

PR #186, opened 2024-03-14
Support alternate audio vizualizers

mbklein:
I did this so that I could plug in [audioMotion-analyzer](https://github.com/hvianna/audioMotion-analyzer) and make use of its full range of options. And it has [_a lot_ of options](https://audiomotion.dev/demo/fluid.html). Sample implementation of an alternative visualizer can be seen [here](https://github.com/mbklein/family-oral-history/blob/main/src/app/components/AudioMotionVisualizer.tsx), and how it integrates with the Viewer [here](https://github.com/mbklein/family-oral-history/blob/main/src/app/page.tsx). It's a bit clunky with all the refs and casting, but I couldn't find a way around it.

Here it is in action, with the same 15-second audio clip (redacted) and three different configurations:

https://github.com/samvera-labs/clover-iiif/assets/196872/da88c1ae-d283-44d4-a3ba-b86e40fa2c9e

https://github.com/samvera-labs/clover-iiif/assets/196872/0813926c-b92c-43ae-98fe-ee4503caac00

https://github.com/samvera-labs/clover-iiif/assets/196872/1da2ff74-2506-433d-b3f0-4cb10989dfba

The AudioMotion implementation could also just be pulled directly into Clover, but it would add another dependency you might not want.

### DIFF
--- a/src/components/Viewer/Player/AudioVisualizerBase.tsx
+++ b/src/components/Viewer/Player/AudioVisualizerBase.tsx
@@ -1,0 +1,4 @@
+import { ForwardRefExoticComponent, RefAttributes } from "react";
+export type AudioVisualizerBase = ForwardRefExoticComponent<
+  RefAttributes<HTMLElement>
+>;
--- a/src/components/Viewer/Player/Player.test.tsx
+++ b/src/components/Viewer/Player/Player.test.tsx
@@ -3,6 +3,7 @@ import { render, screen } from "@testing-library/react";
 
 import { AnnotationResources } from "src/types/annotations";
 import { LabeledIIIFExternalWebResource } from "src/types/presentation-3";
+import AudioVisualizer from "src/components/Viewer/Player/AudioVisualizer";
 import Player from "src/components/Viewer/Player/Player";
 import React from "react";
 import { Vault } from "@iiif/vault";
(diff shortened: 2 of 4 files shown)


mathewjordan:
This is a pretty cool idea. I think we can figure it out but just need to document it.

Would be to cool to see this idea with https://wavesurfer.xyz/

Now that you have spent some time in the code, how do you think we can cleanup the Viewer Context and the config for the better? It seems like we have a highly customizable Viewer here, which is great, but I'm just trying to figure out the best way to let it be malleable without it becoming unwieldy.

mbklein:
wavesurfer looks fun, and it already has a [react hook](https://wavesurfer.xyz/examples/?react.js). It's different from the other visualizers in that it represents the track as a timeline from left to right instead of a changing point-in-time graph – not a problem; just a different approach. I wonder how it would deal with streaming audio not loading all at once, especially for long streams.

I wonder if the context and config could be pluggable? Let different pieces (like the information panel) register themselves and be responsible for their own config/event logic? I don't think it would be that difficult to delegate.

mathewjordan (commented):
Hey @mbklein, I put some eyes on this again. I think we just need to add some docs for these options to push it through -- unless we wanted to tidy up configuration as noted above first.

mbklein:
The behavior this PR was supposed to support can now be done more easily through a combination of Custom Displays and the changes in #284. Closing.

PR #187, opened 2024-03-14
Fix blank page aside rendering bug

adamjarling:
## What does this do?
Quick patch to account for a state where:
- `InformationPanel` open context value is `false`
- "About" in `InformationPanel` configured to be `true`
- No annotation resources exist

It results in the Viewer rendering an empty space where the `InformationPanel` displays (not good)

image

## How to test
With default configuration, run in your dev environment and the initial render should not display empty space to the right of the Viewer Content:

image

### DIFF
--- a/pages/docs/viewer.mdx
+++ b/pages/docs/viewer.mdx
@@ -172,6 +172,7 @@ The information panel is a collapsible panel that displays information about the
 | ---------------------------------------------- | --------- | -------- | ------- |
 | `options.informationPanel.open`                | `boolean` | No       | true    |
 | `options.informationPanel.renderAbout`         | `boolean` | No       | true    |
+| `options.informationPanel.renderAnnotation`    | `boolean` | No       | true    |
 | `options.informationPanel.renderSupplementing` | `boolean` | No       | true    |
 | `options.informationPanel.renderToggle`        | `boolean` | No       | true    |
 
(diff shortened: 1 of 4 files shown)


tpendragon:
Oh I think I fixed this in https://github.com/samvera-labs/clover-iiif/pull/185 by removing the <Aside> container, which could enable some CSS transitions, but whatever y'all think is best.

mathewjordan (changes requested):
@adamjarling This is close but it's not quite correct with Manifests that have annotations (VTT or otherwise). It does behave as expected where it renders the aside, however if the aside is toggled off, the canvas does not grow to take the removed space.

See with:

- VTT: https://digital.lib.utk.edu/assemble/manifest/rfta/118
- Simple Tagging Annotation - https://iiif.io/api/cookbook/recipe/0021-tagging/manifest.json

adamjarling:
> @adamjarling This is close but it's not quite correct with Manifests that have annotations (VTT or otherwise). It does behave as expected where it renders the aside, however if the aside is toggled off, the canvas does not grow to take the removed space.
> 
> See with:
> 
> * VTT: https://digital.lib.utk.edu/assemble/manifest/rfta/118
> * Simple Tagging Annotation - https://iiif.io/api/cookbook/recipe/0021-tagging/manifest.json

@mathewjordan Ok, just pushed up what should be a fix.  Trying to capture this with tests, feels a little brittle.  [Playwright](https://playwright.dev/)/Cypress to the rescue might save some time here in the future.

mathewjordan (approved):

PR #194, opened 2024-04-09
Add Rendering primitive.

markpbaggett:
# What Does This Do

Adds a basic rendering primitive for when you have a `resource that is an alternative, non-IIIF representation of the resource`. This just takes that thing and creates an anchor to the external resource similar to how things work with `SeeAlso`.

### DIFF
--- a/src/components/Primitives/Rendering/Rendering.tsx
+++ b/src/components/Primitives/Rendering/Rendering.tsx
@@ -1,0 +1,39 @@
+import React from "react";
+import { styled } from "src/styles/stitches.config";
+import { getLabelAsString } from "src/lib/label-helpers";
+import { PrimitivesRendering } from "src/types/primitives";
+import { sanitizeAttributes } from "src/lib/html-element";
+
+const StyledRendering = styled("li", {});
+const StyledWrapper = styled("ul", {});
+
+const Rendering: React.FC<PrimitivesRendering> = (props) => {
+  const { as, rendering } = props;
+
+  /**
+   * Create attributes and remove React props
+   */
+  const remove = ["as", "rendering"];
+  const attributes = sanitizeAttributes(props, remove);
+  console.log(rendering);
+  return (
+    <StyledWrapper as={as}>
+      {rendering &&
+        rendering.map((resource) => {
+          const label = getLabelAsString(
+            resource.label,
+            attributes.lang,
+          ) as string;
+          return (
+            <StyledRendering key={resource.id}>
+              <a href={resource.id} {...attributes}>
+                {label ? label : resource.id}
+              </a>
+            </StyledRendering>
+          );
+        })}
+    </StyledWrapper>
+  );
+};
+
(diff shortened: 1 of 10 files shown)


src/components/Primitives/Rendering/Rendering.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const Rendering: React.FC<PrimitivesRendering> = (props) => {
+  const { as, rendering } = props;
+
+  /**
+   * Create attributes and remove React props
+   */
+  const remove = ["as", "rendering"];
+  const attributes = sanitizeAttributes(props, remove);
+  console.log(rendering);
```
mathewjordan, line 18:
```suggestion

```

src/components/Viewer/InformationPanel/About/About.tsx
```diff
@@ -56,6 +59,9 @@ const About: React.FC = () => {
         <SeeAlso
           seeAlso={seeAlso as unknown as PrimitivesExternalWebResource[]}
         />
+        <Rendering
+          rendering={rendering as unknown as PrimitivesExternalWebResource[]}
+        />
```
mathewjordan, line 64: I wonder of this should have a `target="_blank"`. Might be assumptive to force a download but it could be beneficial for the user to not be navigated away here. This might be true for all of these linking properties, honestly.

mathewjordan (changes requested):
Looks good,. Thanks for pushing this forward. We should add some docs in `pages` and include it on the homepage as well.

src/components/Viewer/InformationPanel/About/About.tsx
markpbaggett: Added this here, but I agree it should probably be everywhere.

src/components/Viewer/Properties/Rendering.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+const PropertiesRendering: React.FC<PropertiesRenderingProps> = ({
+  rendering,
+}) => {
+  if (rendering?.length === 0) return <></>;
+
+  return (
+    <>
+      <span className="manifest-property-title">Rendering</span>
```
markpbaggett, line 16: I opened my manifest in some other viewers and started to second guess the label of the span here. Mirador calls this, "Alternate formats." Do you think that is better?  Rendering means nothing to me as a user.

mathewjordan: Yeah, let's copy Mirador here. This should intersect nicely with Adam's work in #190 too once that gets pulled through.

markpbaggett:
@mathewjordan I'm going to go ahead and write docs, then I'll re-request a review.

mathewjordan (approved):
:shipit: Thanks for adding this. Want me to merge, version, and publish?

markpbaggett:
@mathewjordan Sure. Sounds good.

PR #198, opened 2024-04-17
Load HLS manifests that don't end in .m3u8 if the format is right.

### LINKED ISSUE #197: HLS should render if the mime-type is an application/vnd.apple.mpegurl, not just if the file ends with .m3u8
We noticed that if we append parameters (we add an authorization token sometimes) to the .m3u8 that it doesn't render, because it only loads HLS if it ends with an extension rather than going off the "format." Would that be a big lift? Or at least making it so either or would work?

tpendragon:
Closes #197

### DIFF
--- a/src/components/Viewer/Player/Player.tsx
+++ b/src/components/Viewer/Player/Player.tsx
@@ -9,6 +9,7 @@ import { LabeledIIIFExternalWebResource } from "src/types/presentation-3";
 import { PlayerWrapper } from "src/components/Viewer/Player/Player.styled";
 import Track from "src/components/Viewer/Player/Track";
 import { getPaintingResource } from "src/hooks/use-iiif";
+import { hlsMimeTypes } from "src/lib/hls";
 
 // Set referrer header as a NU domain: ie. meadow.rdc-staging.library.northwestern.edu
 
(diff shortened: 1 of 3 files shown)


src/components/Viewer/Player/Player.tsx
```diff
@@ -51,8 +52,14 @@ const Player: React.FC<PlayerProps> = ({
...
      * the IIIF content resource ID is not .m3u8
+     * and format is not one of many m3u8 formats.
      */
-    if (painting.id.split(".").pop() !== "m3u8") return;
+    if (
+      painting.id.split(".").pop() !== "m3u8" &&
+      painting.format &&
+      !hlsMimeTypes.includes(painting.format)
+    )
+      return;
```
mathewjordan, line 62: My brain is slow today, but I wonder if we still need the `&&` to check for `m3u8` in the `id` at all?

```suggestion
    if (!hlsMimeTypes.includes(painting?.format)) return;
```

src/components/Viewer/Player/Player.tsx
```diff
@@ -51,8 +52,14 @@ const Player: React.FC<PlayerProps> = ({
     /**
      * Eject HLS attachment if file extension from
      * the IIIF content resource ID is not .m3u8
+     * and format is not one of many m3u8 formats.
```
mathewjordan, line 55:
```suggestion
     * Eject HLS attachment if format is not one
     * of the allowed m3u8 formats.
```

mathewjordan (changes requested):
This looks great @tpendragon. Thanks for putting this through. It's nice to have y'all in here to review some of our localized oversights.

What do you think of my suggestion? I think we can rely solely on `format` of the resource at this point.

src/components/Viewer/Player/Player.tsx
tpendragon: We can do that, it's just backwards incompatible. Like if someone didn't have that format before it would have worked.

mathewjordan: Very true. I'm unaware of anyone else using Clover with HLS but I cases could be out there.

mathewjordan (approved):
👍 Let's err on the safe side and roll with your initial changes.

Issue #203, opened 2024-05-01
Clover doesn't always toggle between canvases on thumbnail click


mathewjordan:
## Description

Noted by @wykhuh 

> One issue I noticed is there is a intermittent bug  where the  image shown in the image viewer  sometimes does not change when the canvas is updated using  updateActiveCanvas

```
      dispatch({
        type: "updateActiveCanvas",
        canvasId: canvasId,
      });

```

> For  instance, I was on the third image, and I clicked on the second thumbnail, but the image viewer  still shows the third image while the thumbnail bar shows the second image as being active. This bug only happens sometimes. 

## Done Looks Like

- [ ] Clover reliably updates active canvas

PR #204, opened 2024-05-01
Force redraw to destroy previous tiled image.

mathewjordan:
This addresses issue #203 raised by @wykhuh. In some instances when navigating canvases of a manfiest, the expected IIIF tiled images is not drawn to the canvas and the previous item is retained. The actual occurance is a collision on the index of the canvas where the IIIF tiled image is strangely drawn under the previous tiled image that was somehow retained. In addition to the redraw, we now also update a timing issue where the `osdUri` is not reset unless the `openSeadragon` instance exists.

To resolve, we are now forcing a redraw prior to the drawing of the new image.

This also updates the `openseadragon` dependency as previous IIIF tile size issues have now been resolved in the recent release 4.1.1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "flexsearch": "^0.7.43",
     "hls.js": "^1.5.3",
     "node-webvtt": "^1.9.4",
-    "openseadragon": "^2.4.2",
+    "openseadragon": "^4.1.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-error-boundary": "^4.0.12",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -83,7 +83,7 @@
     "flexsearch": "^0.7.43",
     "hls.js": "^1.5.3",
     "node-webvtt": "^1.9.4",
-    "openseadragon": "^2.4.2",
+    "openseadragon": "^4.1.1",
```
adamjarling, line 86: Nice

adamjarling (approved):
Good catch on the bug.

Issue #205, opened 2024-05-01
Tiled image not loading correctly


wykhuh:
I create two  new apps using both Next.js and React + Vite.js, and installed Clover 2.8.1. In both cases, the tiled image is not working.  It shows a small thumbnail, but we can't zoom the image.

<img width="713" alt="image" src="https://github.com/samvera-labs/clover-iiif/assets/6968611/522a7c60-46b2-4764-8c5d-d9e54b63db20">

I then uninstalled Clover, and installed previous versions of clover. version 2.5 works in both apps, which is before adding the image viewer component.

Nextjs repo https://github.com/wykhuh/clover-nextjs-demo

```jsx
import dynamic from "next/dynamic";

const Viewer = dynamic(
  () => import("@samvera/clover-iiif").then((Clover) => Clover.Viewer),
  {
    ssr: false,
  },
);

export default function Home() {
  const iiifContent =
    "https://api.dc.library.northwestern.edu/api/v2/collections/c373ecd2-2c45-45f2-9f9e-52dc244870bd?as=iiif";

  return <Viewer iiifContent={iiifContent} />;
}
```

React repo  https://github.com/wykhuh/clover-react

```jsx
import Viewer from "@samvera/clover-iiif/viewer";

function App() {
   const iiifContent =
    "https://api.dc.library.northwestern.edu/api/v2/collections/c373ecd2-2c45-45f2-9f9e-52dc244870bd?as=iiif";

  return <Viewer iiifContent={iiifContent} />;
}

export default App
```

PR #208, opened 2024-05-01
Initialize new openseadragon instance once.

mathewjordan:
This addresses #205, caused by an issue in React strict mode where new openseadragon instances are opened on each re-render. This work sets a ref to initialize the the component, and then OSD.tsx will not instantiate a `new OpenSeadragon.Viewer()` if this ref is `true`.

This issue is best tested external to the Nextra docs.

### DIFF
--- a/src/components/Image/OSD/OSD.tsx
+++ b/src/components/Image/OSD/OSD.tsx
@@ -4,7 +4,7 @@ import {
   Wrapper,
 } from "src/components/Image/Image.styled";
 import OpenSeadragon, { Options } from "openseadragon";
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 
 import Controls from "src/components/Image/Controls/Controls";
 import { OpenSeadragonImageTypes } from "src/types/open-seadragon";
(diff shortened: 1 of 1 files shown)


adamjarling (approved):
@mathewjordan This does the trick, testing locally in a `link`ed project.   Now that I see this, wondering what would happen if we bypassed putting the openSeadragon instance on state at all, and put your `ref` on the instantiation?   Kinda like:

```
const osdRef = useRef<OpenSeadragon.Viewer>();

useEffect(() => {
  if (!osdRef.current) {
    osdRef.current = OpenSeadragon(config);
  }
}, [config]);
```

And then swapped out `openSeadragon` with `osdRef.current` everywhere? 

Either way I think this arrives at the right destination (isolating the OpenSeadragon instance from React renders), via `useRef()`.  👍

mathewjordan:
> @mathewjordan This does the trick, testing locally in a `link`ed project. Now that I see this, wondering what would happen if we bypassed putting the openSeadragon instance on state at all, and put your `ref` on the instantiation? Kinda like:
> 
> ```tsx
> const osdRef = useRef<OpenSeadragon.Viewer>();
> 
> useEffect(() => {
>   if (!osdRef.current) {
>     osdRef.current = OpenSeadragon(config);
>   }
> }, [config]);
> ```
> 
> And then swapped out `openSeadragon` with `osdRef.current` everywhere?
> 
> Either way I think this arrives at the right destination (isolating the OpenSeadragon instance from React renders), via `useRef()`. 👍

This looks solid but, I attempted the approach and ran into an issue where OpenSeadragon isn't liking being in captured in the ref. Did you get it working on your end @adamjarling?

PR #218, opened 2024-06-06
WebVTT: support styled cues and voice spans

mbklein:
This PR adds support for [advanced WebVTT features](https://www.speechpad.com/captions/webvtt#toc_7) such as text formatting, CSS styling, voice spans, and Ruby characters.

To do so, I had to switch VTT parsing from [`node-webvtt`](https://github.com/osk/node-webvtt) to [`vtt.js`](https://github.com/mozilla/vtt.js). `vtt.js` is older, but neither of them has seen significant development in years, so I'm going with the theory that the lack of activity means more "stable" and less "abandoned."

The main advantage of this approach is access to the `WebVTT.convertCueToDOMTree()` method, which converts each WebVTT cue to a DOM `DocumentFragment` according to the guidelines laid out in [Section 6.5 of the WebVTT W3C recommendation](https://www.w3.org/TR/webvtt1/#dom-construction-rules). This allows the consuming application to style cues according to class or voice in addition to stripping out any other tags that shouldn't be there.

Behavior for untagged cues is unchanged.

#### Cue
```webvtt
00:00:02.320 --> 00:00:04.062
<v interviewer>How did you meet your wife?</v>
```

#### Before
```html
&lt;v interviewer&gt;How did you meet your wife?&lt;/v&gt;
<strong>0:02</strong>
```
SCR-20240606-nyki

#### After (with styling applied to the resulting `.webvtt-cue [title="..."]` selectors)
```html
<div class="webvtt-cue">
  <span title="interviewer">How did you meet your wife?</span>
</div>
<strong>0:02</strong>
```
SCR-20240606-nxii


src/hooks/use-webvtt.ts
```diff
@@ -89,11 +93,40 @@ const useWebVtt = () => {
...
+
+        cues.push({
+          identifier: uuidv4(),
+          start: cue.startTime,
+          end: cue.endTime,
+          align: cue.align,
+          html,
+          text,
+        });
+      };
```
mathewjordan, line 116: I'm seeing some initial issues where cues are not rendering as expected. Something in here seems to be not parsing correctly on the lunchroom manners recipe. 

image

mathewjordan: The VTT definitely loads and we get cues/time(s), but the text is coming back empty.

mathewjordan (changes requested):
I'm all for swapping out. It looks like it fails on 
https://iiif.io/api/cookbook/recipe/0219-using-caption-file/manifest.json

Can we also add an example manifest with a styled VTT resource somewhere in `public/manifest`? You can host the styled VTT in public as well.

PR #222, opened 2024-07-10
Align the download icon.

tpendragon:
This goes from

Screen Shot 2024-07-10 at 10 47 00 AM

to

Screen Shot 2024-07-10 at 10 46 54 AM

for us


mathewjordan (approved):
Awesome, thanks for the fix.

PR #228, opened 2024-10-10
Fixes bug in copy manifest url

kdid:
## Summary

- Fixes bug with "Copy Manifest URL" link where "Failed" is eventually displayed after a successful copy and "Success" message
- Clear copy status text in `useCopyToClipboard` `useEffect`  after timeout rather than displaying a failure message


mathewjordan (approved):
🚀

PR #232, opened 2024-11-11
Set default labels as index + 1 for canvas figures.

mathewjordan:
## What does this do?

This sets the default label for canvas figures (thumbnails) in the media component in cases where the Canvas does not have a label. The label will be the `(index + 1).toString()` to accommodate all languages. This same label will be the `alt` of the rendered thumbnail as well. Filtering will also work in this cases, ex: "2" as the input value will be show the "2", "12", "21", "22", etc... thumbnails.

## How to review?

Review with this fixture
https://samvera-labs.github.io/clover-iiif/docs/viewer/demo?iiif-content=https%3A%2F%2Fmarkpbaggett.github.io%2Fstatic_iiif%2Fmanifests%2Ftests%2Fservice_maps_0103.json

 - Canvas thumbnails will render as "1" and "2'.
 - Canvases are filterable

### DIFF
--- a/src/components/Viewer/Media/Media.test.tsx
+++ b/src/components/Viewer/Media/Media.test.tsx
@@ -1,12 +1,53 @@
+import { render, screen } from "@testing-library/react";
+
+import { Canvas } from "@iiif/presentation-3";
 import Media from "src/components/Viewer/Media/Media";
 import React from "react";
-import { render, screen } from "@testing-library/react";
+
+const items = [
+  {
+    id: "https://example.org/manifest/1/canvas/1",
+    type: "Canvas",
+    height: 600,
+    width: 800,
+    label: { none: ["Canvas 1"] },
+    thumbnail: [
+      {
+        id: "https://example.org/item/1/default.jpg",
+        type: "Image",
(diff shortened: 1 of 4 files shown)


markpbaggett (commented):
Looks good.  I noticed one thing:

If you try search via filter, things work as expected except when the filter is applied the label of the Canvas always switches to `1` even if you search for `2` and it finds `2`.  Not sure you care about this, but I thought I'd mention. Looks good other than that.  Thanks again.

mathewjordan:
> Looks good. I noticed one thing:
> 
> If you try search via filter, things work as expected except when the filter is applied the label of the Canvas always switches to `1` even if you search for `2` and it finds `2`. Not sure you care about this, but I thought I'd mention. Looks good other than that. Thanks again.

Oh, great catch. This update should catch that:

https://github.com/samvera-labs/clover-iiif/blob/7d658836f81af6d76ee9fa32f3e84a622accb83b/src/components/Viewer/Media/Media.tsx#L124

markpbaggett (approved):
Looks good!  :shipit:

PR #239, opened 2024-12-12
Detect dir of rtl or ltr based on BCP47 language codes

kdid:
- Determine direction of language from the BCP47 code.


mathewjordan (approved):
🎸

Issue #245, opened 2025-01-29
Viewer - Custom "Loading" Component


joseph-bayer:
I'd like to display a custom component instead of the "Loading" text that displays while the manifest is fetched & loaded into React.Context ([code here](https://github.com/samvera-labs/clover-iiif/blob/b33af20b375e522ab750ba8047669cf0b4fdded7/src/components/Viewer/index.tsx#L192))

1. Is this something you'd be interested in me implementing?
2. If so, what solution would be acceptable? For example, we could pass in a react component into the config and display that instead of "Loading" (but only when that config value is populated)

EDIT: Support for a custom loading component is now available! Based on the conversation in [this pull request](https://github.com/samvera-labs/clover-iiif/pull/246), I'll be updating the default loading text to be a skeleton instead.

mathewjordan:
Hey @joseph-bayer this would be a great addition. That `Loading` text has always bothered me a bit. Updating this default component and setting up a way to pass in a custom React component through the configuration options seems like a great step forward.

We're open to PRs and if our configuration models make sense and you have thoughts on how this could work, please contribute forward!

mathewjordan:
@joseph-bayer I'll leave it up to whether to amend this issue or create a new one for the update of the default loader using a skeleton.

joseph-bayer:
Hey @mathewjordan - we can amend this issue for the loading skeleton.

Sorry for the delay! I'm still working on it, but have been super busy.

PR #246, opened 2025-01-30
Add support for a custom loading component

joseph-bayer:
Hey there!

This is a pull request to add support for replacing the default loading UI ("Loading") with a React component. The idea is from this issue: https://github.com/samvera-labs/clover-iiif/issues/245

I had to use React.ComponentType as opposed to the more flexible ReactNode because ReactNode caused an infinite loop in the `deepMerge` function.

In the [issue thread](https://github.com/samvera-labs/clover-iiif/issues/245), it was mentioned that a new default loading state would be nice. I'm open to implementing that! For example, we could do [some sort of skeleton](https://ui.shadcn.com/docs/components/skeleton). If we do implement a skeleton, let me know how detailed it should be (big rectangle vs lots of smaller shapes).

### DIFF
--- a/pages/docs/viewer.mdx
+++ b/pages/docs/viewer.mdx
@@ -494,7 +494,7 @@ return (
 
 ### Custom Loading Component
 
-You may choose to override the default loading component that displays while the manifest is fetched and loaded into state.
+Clients may choose to override the default _loading_ component that displays while the IIIF resource is fetched and loaded into state.
 
 ```jsx
 <Viewer
--- a/pages/docs/viewer.mdx
+++ b/pages/docs/viewer.mdx
@@ -492,6 +492,21 @@ return (
 
 ---
 
+### Custom Loading Component
+
+You may choose to override the default loading component that displays while the manifest is fetched and loaded into state.
+
+```jsx
+<Viewer
+  iiifContent="https://example.com/manifest/1"
+  options={{
+    customLoadingComponent: () => <>My custom loading component</>,
+  }}
+/>
+```
+
+---
+
 ### Custom canvas displays
 
 Clients may wish to use their own display components instead of Clover Viewer's default displays ([OpenSeadragon](https://openseadragon.github.io/) for images and [HTML Video Player](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video) for audio/video). The `Viewer` component allows a client to target individual `canvas` items in a IIIF Manifest by either direct reference to a canvas `id` or `format` (ie. `video/ogg`). See the Type Definition below for `CustomDisplay`, and an example implementation.
--- a/pages/docs/viewer.mdx
+++ b/pages/docs/viewer.mdx
@@ -142,27 +142,28 @@ const MyCustomViewer = () => {
 
 `Viewer` can configured through an `options` prop, which will serve as a object for common options.
 
-| Prop                            | Type                                                                        | Required | Default                                  |
-| ------------------------------- | --------------------------------------------------------------------------- | -------- | ---------------------------------------- |
-| `iiifContent`                   | `string`                                                                    | Yes      |                                          |
-| `iiifContentSearchQuery`        | [See Content Search](#content-search)                                       | No       |                                          |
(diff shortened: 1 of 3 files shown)


mathewjordan:
Hey @joseph-bayer  I'm going to review this as-is but perhaps we can still work towards updating the default with a skeleton? I image something very simple that grows and conforms with the general layout of Clover. The only large rectangle would be the canvas where images/video are rendered, something giving the impression of rich content.

██████████████

███████████████████████   █████████
███████████████████████   █████  
███████████████████████   ███████   
███████████████████████   ████████ 
███████████████████████
███████████████████████

Related but not blocking changes is getting Clover fully working with Radix UI Themes which has a Skeleton component https://www.radix-ui.com/themes/docs/components/skeleton. This work is being completed in https://github.com/samvera-labs/clover-iiif/pull/202 but is a bit stale right now.

mathewjordan (changes requested):
This looks great and functional. Thanks for addition.

Can we also add a section in the docs in `viewer.mdx` for with a simple example?

```jsx
<Viewer
  iiifContent="https://example.com/manifest/1"
  options={{
    customLoadingComponent: () => <>My custom loading component</>,
  }}
/>
```

joseph-bayer:
> Hey @joseph-bayer I'm going to review this as-is but perhaps we can still work towards updating the default with a skeleton? I image something very simple that grows and conforms with the general layout of Clover. The only large rectangle would be the canvas where images/video are rendered, something giving the impression of rich content.
> 
> ```
> ██████████████
> 
> ███████████████████████   █████████
> ███████████████████████   █████  
> ███████████████████████   ███████   
> ███████████████████████   ████████ 
> ███████████████████████   
> ███████████████████████
> ```
> 
> Related but not blocking changes is getting Clover fully working with Radix UI Themes which has a Skeleton component https://www.radix-ui.com/themes/docs/components/skeleton. This work is being completed in #202 but is a bit stale right now.

I think this is totally doable. We could even add/remove certain elements of the skeleton based on the config (ex: if the information panel is closed by default).

I you'd like, I could implement a solution independent from Radix (just pure html + css), and leave a TODO: comment in the code to utilize the Radix Skeleton component once themes are implemented.

Should I create a separate issue + pull request for this?

pages/docs/viewer.mdx
```diff
@@ -491,6 +492,21 @@ return (
 
 ---
 
+### Custom Loading Component
+
+You may choose to override the default loading component that displays while the manifest is fetched and loaded into state.
```
mathewjordan, line 497:
```suggestion
Clients may choose to override the default _loading_ component that displays while the IIIF resource is fetched and loaded into state.
```

mathewjordan (approved):
🌟

mathewjordan:
> > Hey @joseph-bayer I'm going to review this as-is but perhaps we can still work towards updating the default with a skeleton? I image something very simple that grows and conforms with the general layout of Clover. The only large rectangle would be the canvas where images/video are rendered, something giving the impression of rich content.
> > ```
> > ██████████████
> > 
> > ███████████████████████   █████████
> > ███████████████████████   █████  
> > ███████████████████████   ███████   
> > ███████████████████████   ████████ 
> > ███████████████████████   
> > ███████████████████████
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Related but not blocking changes is getting Clover fully working with Radix UI Themes which has a Skeleton component https://www.radix-ui.com/themes/docs/components/skeleton. This work is being completed in #202 but is a bit stale right now.
> 
> I think this is totally doable. We could even add/remove certain elements of the skeleton based on the config (ex: if the information panel is closed by default).
> 
> I you'd like, I could implement a solution independent from Radix (just pure html + css), and leave a TODO: comment in the code to utilize the Radix Skeleton component once themes are implemented.
> 
> Should I create a separate issue + pull request for this?

Yes please. New issue is fine as well. But a seperate PR makes sense. I'll send this through. Thanks again! Best.

mathewjordan:
This work has been versioned as `@samvera/clover-iiif@2.12.2`

PR #256, opened 2025-03-28
Refine information panel toggle, address responsive issues.

mathewjordan:
## Description

This gem addresses a responsive issues relating to legacy ideas for the Information Panel component. Previously, the information panel was treated very differently in smaller view ports, forcing the viewer to go to `position: fixed`. This is no longer desired as it grants Clover too much control over the screen within a consuming application. With the @tpendragon and Princeton's need of Clover to reside within a relative container wrapping it if `options.canvasHeight: "100%"` is set, this convention presents additional complications.

## What does this do?

- This work advances Clover by treating the Toggle the same regardless of viewport size. 
- In `@sm` viewports, the expanded Information Panel `aside` now slides completely over the viewer, overtaking width and height of the the content space provided.
- In cases where the `informationPanel.open` `informationPanel.renderToggle` are both false, the Information Panel should not rendered by default, or be able to be toggled as rendered.
- The exception to the above cases is if Annotations are present, and the informationPanel.defaultTab has been set to a custom value of `manifest-annotations`, as is the case: https://github.com/nulib/dc-nextjs/blob/deploy/staging/components/Clover/ViewerWrapper.tsx#L59-L64
- I have also added a **Viewer** tab that is only visible on smaller viewports, making it more obvious to users that may want to toggle back to the relative Image, Video, or Sound resources.

**Larger viewport, open**

image

**Larger viewport, closed**

image

**Small viewport, open**

image

**Small viewport, closed**

image

### DIFF
--- a/pages/docs/viewer.mdx
+++ b/pages/docs/viewer.mdx
@@ -19,15 +19,15 @@ A UI component that renders a multicanvas IIIF item viewer with pan-zoom support
 
 ---
 
-<Viewer
-  iiifContent="https://api.dc.library.northwestern.edu/api/v2/works/8a833741-74a8-40dc-bd1d-c416a3b1bb38?as=iiif"
-  options={{
-    showIIIFBadge: false,
-    informationPanel: {
-      open: false,
-    },
-  }}
-/>
+<div style={{ position: "relative", height: "500px", zIndex: "0" }}>
+  <Viewer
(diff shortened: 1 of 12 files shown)


charlesLoder (approved):
Looks great, so much cleaner!

tpendragon:
@mathewjordan Thank you!

PR #266, opened 2025-04-24
Consolidate Scroll figure component. 

muildrik:
In this PR we let FigureImageViewer return a thumbnail, if that is available in the manifest; otherwise we default to what's in the body property. Using the thumbnail greatly speeds up loading time since thumbnail images are typically a lot smaller. This means that we can reduce two components to one. I've updated the unit tests to reflect this as well.

In a follow-up PR I plan to implement an intersect/observer on the items array to keep tabs on what's visible and update openSeadragon (OSD.tsx) to lazy-load whatever is on screen. The thumbnail image should then be returned by default until the first one or two levels of the pyramid are available so that the user can start to interact with the actual IIIF data on the canvas.


mathewjordan (approved):
Looks solid. Thanks for cleaning this up.

PR #267, opened 2025-04-28
Make sure the manifest has a service property, and then check manifes…

muildrik:
This checks to see if a `service` property is defined before assuming it to be an array.


mathewjordan (approved):
👍

PR #268, opened 2025-04-28
Hotfix: show interior scrollbars only when necessary

muildrik:
## What does this do?

Adjusts overflow behavior to show scrollbar only when content is clipped.


mathewjordan (approved):
👍

PR #273, opened 2025-05-05
Added LazyLoad component to the ScrollItems in Scroll component

muildrik:
This seems to do the job on a manifest with 188 pages. The only issue seems to be that when the user loads the page kind of mid-way the LazyLoad component is not aware of the position of the user and gets a little behind loading what is on the screen. We may need to add something there to check what will be on the screen as the page itself is loading.


mathewjordan (approved):
Love it.

mathewjordan:
@muildrik Thanks for adding! I agree there are still some issues with timing the load in certain situations but overall this greatly helps initial pageload. Let's itemize edge cases and address in iterations.

PR #275, opened 2025-05-06
Use @iiif/helpers for thumbnail generation.

mathewjordan:
## What does this do?

This work does several things to address issues related to thumbnail generation for canvases in the Viewer component.

Three highlighted items addressed:

- An issue with URL encoding of `%` characters in Vault within `@iiif/helpers/vault` was causing a double-encoding problem with URLs related Princeton IIIF Image services.
- Even if this thumbnail retrieval succeeded, legacy code related to thumbnails was often retrieving very large images and not utilizing image services.
- The `info.json` retrieval was not being completed behind a lazy load mechanism.

### Solution

- Deletion of homegrown thumbnail generation helper and opting for use of `getThumbnail()` in the `@iiif/helpers/thumbnail`. This helper seeks out the "best" possible candidate thumbnail for a respective canvas and returns it.
- The helper is only called if the thumbnail wrapping item has been lazyloaded. In this, two possible requests are made. One to the `info.json` of the image service, and the second the URI for the `default.jpg`
- By default the Thumbnail component will also now seek out `thumbnail` property first. 

## Result

Northwestern (thumbnails on canvas)

image

Princeton (thumbnails generated from image services)

image

image

Bodleian (thumbnails generated from image services)

image


mathewjordan:
Hey @AminEdZare and @tpendragon I believe we have addressed this issue. I opted for a bit of an overhaul of thumbnail generation rather than seeking to bandaid with `decodeURIComponent()`. I could not request @AminEdZare for some reason (maybe samvera labs user related).

tpendragon (approved):
Looks amazing! Thank you!

mathewjordan:
@tpendragon Cool, I'll version and release.

PR #290, opened 2025-06-24
Add ability to accept prefetched JSON as `iiifContent`

mathewjordan:
## What does this do?

This work adds the ability for the `iiifContent` prop to accept IIIF Manifests and Collections as JSON objects. In some use cases applications may already have these resources ready to feed to a viewer and do not want to make additional requests on the the `id`. This work accounts for manual upgrading of Presentation 2.x resources to 3.0 if required. It then sideloads the resource into `vault`. 

## How should we review?

If spinning up a Clover locally, navigate to the hidden page of http://localhost:3000/docs/viewer/prefetched. On this page, a resource will be loading. Note that no additional XHR request is made on the Manifest.

### DIFF
--- a/pages/docs/viewer.mdx
+++ b/pages/docs/viewer.mdx
@@ -15,6 +15,45 @@ export const handleContentStateCallback = ({ encoded, json }) => {
   });
 };
 
+export const prefetchedManifestJson = {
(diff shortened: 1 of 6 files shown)


tpendragon (approved):
Very cool! Thank you!

PR #307, opened 2025-10-15
Refactor panel visibility logic

### LINKED ISSUE #306: Refactor InformationPanel visibility logic to allow sidebar when any tab is present (including plugins)
I have a chunk of time to work on OSS stuff, so I figured I'd tackle this minor annoyance. This issue prevents us from fully disabling the "About" panel (while keeping our Table of Contents structures plugin) in our DC app.

Currently, the InformationPanel sidebar only opens when `renderAbout` is true, or in a narrow annotation-specific "force open" case. If `renderAbout` is false - even with other tabs (like annotation or plugin panels) configured - the panel does not open and is non-functional.

### Steps to Reproduce

- Set `renderAbout: false` in the configuration.
  - The toggle button shows, but no panel opens.
- Set `renderAbout: false`, `renderAnnotation: true`, and provide a manifest with annotations.
  - The panel still does not open, unless `renderToggle: false` (force open condition).
- Add plugin panels with `renderAbout: false`.
  - The sidebar won't open, even if plugins provide tabs.

### Root Cause

Panel visibility in the Viewer is currently gated by the following logic:
```typescript
const isForcedAside =
  hasAnnotations && informationPanel?.renderAnnotation && informationPanel?.renderToggle === false && isInformationOpen;

const isAside =
  (informationPanel?.renderAbout && isInformationOpen) || isForcedAside;
```
- The sidebar only opens if `renderAbout` is true and toggled open, or if the specific force annotation case is met.
- Plugins and other tabs do not affect panel visibility.

### Desired Behavior

- The InformationPanel sidebar should open if **any** panel (About, Annotation, Content Search, Plugin) is configured and available.
- `isAside` should depend on the presence of any configured panel, not just `renderAbout`.
- The "force open" logic (`isForcedAside`) should consider all available panels, not just annotation.

body-clock:
Closes #306

## Summary

This PR refactors InformationPanel visibility and tab-selection logic so the sidebar opens whenever any configured panel tab is actually available, instead of being gated primarily by the About tab.

The original issue was that `renderAbout: false` could make the information sidebar unusable even when annotation tabs, content search results, or plugin-provided panel tabs were configured. This change centralizes panel availability checks and makes `Viewer/Content` decide whether the sidebar has anything renderable before opening it.

## Scope of changes

- Centralizes panel/tab availability in `src/lib/information-panel-helpers.ts`:
  - `hasAnyPanel` determines whether the sidebar has any renderable content.
  - `getAvailableTabs` derives the current tab list from About, annotations, content search, and plugin panels.
  - `getDefaultTab` chooses the configured default when available, otherwise falls back to the first available tab.
  - `annotationTargetsCanvas` handles content-state targets through direct targets, `SpecificResource.source`, string ids, and target arrays.
- Updates `Viewer/Content` so sidebar visibility is based on any available panel:
  - About panel availability.
  - Annotation resources, content-state annotations, or annotation collections.
  - Content search resources.
  - Plugin-provided information panels.
- Moves plugin information-panel discovery up into `Viewer/Content` and passes `pluginsWithInfoPanel` into `InformationPanel`, avoiding duplicate plugin setup and allowing plugin tabs to participate in the top-level sidebar visibility decision.
- Decomposes `InformationPanel` into smaller pieces:
  - `TabList` for tab triggers and the close control.
  - `TabContent` for built-in About, annotation, content-state, annotation collection, and content-search panel content.
  - `PluginTabContent` for plugin-provided panel rendering.
- Extracts InformationPanel hooks:
  - `useFilteredAnnotations` for motivation-aware annotation filtering.
  - `useAvailableTabs` for memoized available-tab calculation.
  - `useTabSelection` for default tab selection and fallback behavior.
- Updates the local dev fixture to exercise a configuration with About disabled and annotation/content-search panels enabled.

## Behavior changes

- The InformationPanel sidebar can open when `renderAbout` is `false`, as long as another tab is available.
- Plugin-provided information panels now count as available sidebar content.
- Annotation tabs are available when annotation resources exist, a matching content-state annotation exists, or an annotation collection has pages.
- Content-state panel checks account for all currently visible canvases, which fixes paged spreads where the target canvas is not the first visible canvas.
- `defaultTab` is honored only when that tab is available.
- If the configured default tab is async, such as content search or annotations, the panel can initially fall back to another available tab and then switch to the configured default once it becomes available.
- Current tab selection is preserved when it remains valid, and the panel falls back only when the current tab is no longer available.

## Review follow-up included

- Fixed declaration-generation build failures from unsafe content-state target typing.
- Broadened content-state target parsing to avoid unsafe casts and support IIIF target shape variations.
- Fixed paged content-state sidebar visibility by checking every visible canvas, not only the first.
- Fixed async default-tab selection so configured defaults are not permanently ignored after an initial fallback.

## Test coverage

Adds and expands coverage for:

- `hasAnyPanel`, `getAvailableTabs`, `getDefaultTab`, and `annotationTargetsCanvas`.
- InformationPanel initial tab selection, reactive tab updates, fallback behavior, async default tabs, and invalid current-tab recovery.
- Viewer content sidebar rendering for About, annotations, content-state annotations, annotation collections, content search, and plugin tabs.
- Paged visible-canvas content-state targeting.

## Verification

Passed locally:

```bash
npm run test:ci -- src/lib/information-panel-helpers.test.ts src/components/Viewer/Viewer/Content.test.tsx src/components/Viewer/InformationPanel/InformationPanel.test.tsx
npm run build
```

Note: `npm run typecheck` currently reports existing test fixture typing issues across several test files, mostly around normalized IIIF fixture shapes and required test props. This PR does not claim a clean full typecheck run until those fixture typings are addressed.


src/components/Viewer/InformationPanel/InformationPanel.tsx
```diff
@@ -116,42 +114,41 @@ export const InformationPanel: React.FC<NavigatorProps> = ({
...
+				informationPanelResource: informationPanel.defaultTab,
+			});
+		} else if (availableTabs.length > 0) {
+			// Otherwise, use the first available tab
+			dispatch({
+				type: "updateInformationPanelResource",
+				informationPanelResource: availableTabs[0],
+			});
+		} else {
+			// If no tabs are available, just default to 'about' - perhaps this should be different?
```
body-clock: Wondering about this - is it worth falling back to a tab that just says something like "No tabs configured, consider adding one or disabling the toggle". Or maybe just disabling the toggle in this case?

PR #333, opened 2026-04-27
Add contentSearchCallback prop on viewer and pre-populate search input with initial query

kdid:
### Summary

Provide a callback so implementing apps can take action when content search query changes. Also fix bug where initial search query is not populating content search field.

### Specific changes

- Pre-populate the search input with `iiifContentSearchQuery.q` on mount. The query was already used to fire the initial search, but the input field showed blank, leaving users with no indication of what had been searched.     
- Added optional `contentSearchCallback?: (query: string) => void` prop to the top-level Viewer. Fires on every keystroke in the search input,  allowing consuming applications to react to query changes. Backwards compatible - no-op when not provided.


mathewjordan (approved):
🔥

PR #343, opened 2026-05-29
fork iiif to forward web component options to react

jeffwklein:
- little fork branch that allows us to use options API through clover-viewer in our web components (still just top level). this lets us pass through the json options outlined in clover react docs: https://samvera-labs.github.io/clover-iiif/docs/viewer#api-reference - there may be some parsing quirks to sort out but it does the trick for us! 

cc @mathewjordan ! just a fun first look at the simplest route to get options passed through from the outer web component layer reliably through to the underlying preact widget

btw- this is only a fork since i didn't wanna bug you or assume i have permission to push to the repo directly. we can talk more about that as we establish a good workflow. 

thanks for all the help mat, can't wait to meet ya in the beautiful netherlands next week!

### DIFF
--- a/src/web-components/clover-viewer.tsx
+++ b/src/web-components/clover-viewer.tsx
@@ -2,9 +2,11 @@ import React, { useLayoutEffect, useRef } from "react";
 
 import Viewer from "src/components/Viewer";
 import register from "src/lib/preact-custom-element/preact-custom-element";
+import { ViewerConfigOptions } from "src/context/viewer-context";
 
 interface CloverViewerAttributes {
   id: string;
+  options?: ViewerConfigOptions;
   /** IIIF Manifest/Collection URL or content state */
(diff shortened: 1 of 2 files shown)


src/web-components/clover-viewer.tsx
```diff
@@ -19,7 +21,11 @@ function CloverViewerWebComponent(
   props: CloverViewerWCProps & CloverViewerAttributes,
 ) {
   const webComponent = useRef<HTMLElement>();
-  const { id } = props;
+  const { id, options } = props;
+
+  const parsedOptions = JSON.parse(options as string);
+  console.log("OPTIONS =====> ", parsedOptions);
```
mathewjordan:
```suggestion
```

mathewjordan (approved):
Let's roll.

PR #351, opened 2026-08-04
Name the image viewport in the Viewer

lukew-cogapp:
`OSD` renders its viewport with `role="img"` and applies an `aria-label` when it gets one:

```tsx
<Viewport
  id={config.id}
  role="img"
  {...(ariaLabel && { "aria-label": ariaLabel })}
/>
```

`Image` derives that label from its `label` prop, which works when you use the component directly. But `Painting` renders `<ImageViewer>` without passing `label`, so inside the packaged `Viewer` the label is always undefined and the viewport ends up as `role="img"` with no accessible name.

The practical effects: axe reports a `role-img-alt` violation, and since OpenSeadragon also makes the canvas focusable, keyboard users reach a tab stop that announces nothing.

## What this changes

Passes a label through from `Painting`:

```tsx
label={imageBody[0]?.label ?? normalizedCanvas?.label ?? undefined}
```

The painting annotation body's label comes first, falling back to the canvas label. In my manifests the bodies have no label but the canvases do (`Image 1`…`Image n`), which isn't inspiring but is a great deal better than nothing. Where neither exists the behaviour is unchanged, and the attribute is still omitted rather than set to an empty string.

## Checks

`npx vitest run` passes (308 tests) and `npm run build` succeeds. Added a test to `Image/index.test.tsx` covering an `InternationalString` label reaching the viewport's `aria-label`, since the existing `OSD.test.tsx` only covers the string case one level down.

Two things I'd welcome your view on:

- Is the body-then-canvas order the one you'd want, or should the canvas label win?
- Would you rather this were opt-in, or configurable text? I've kept it unconditional on the basis that a missing accessible name is a bug rather than a preference, but it does mean viewers with unhelpful canvas labels start announcing them.

### DIFF
--- a/src/components/Image/index.test.tsx
+++ b/src/components/Image/index.test.tsx
@@ -101,4 +101,22 @@ describe("Image component", () => {
       />,
     );
   });
+  it("names the viewport from an InternationalString label", () => {
+    render(
+      <CloverImage
+        body={{
+          id: "https://example.org/image/full/600,/0/default.jpg",
+          type: "Image",
+          format: "image/jpeg",
+          width: 600,
+          height: 600,
+        }}
+        label={{ en: ["Front cover"] }}
+      />,
+    );
+
+    expect(
+      screen.getByTestId("clover-iiif-image-openseadragon-viewport"),
+    ).toHaveAttribute("aria-label", "Front cover");
+  });
 });
--- a/src/components/Viewer/Painting/Painting.test.tsx
+++ b/src/components/Viewer/Painting/Painting.test.tsx
@@ -138,17 +138,17 @@ describe("Painting component", () => {
   it("displays the default Image viewer for non-media files", async () => {
     await vault.loadManifest("", manifestImage);
 
-    const props = {
-      ...defaultProps,
-      activeCanvas:
-        "https://api.dc.library.northwestern.edu/api/v2/works/71153379-4283-43be-8b0f-4e7e3bfda275?as=iiif/canvas/access/2",
-    };
+    const canvasId =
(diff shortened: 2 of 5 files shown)


lukew-cogapp:
Draft @mathewjordan with a couple of questions for you

mathewjordan:
Sorry, I'm a bit hazy after returning from out of the office.

1) I think it would be best to iterate up in granularity. Body -> Annotation -> Canvas. We need to be careful to be aware of this and perhaps compound the array of items at any chosen level into a string `[n]` bodies, or `[n]` canvas as multiple could be present in the viewport. This could be a long Scroll built from multiple canvases or a multi-spectral image with multiple choice bodies. 

2) For most part, Clover always opts toward trusting in the IIIF spec and how the provider of a resource has chosen to describe their content. If the item is not well described in the Manifest, then we could just fill with a generic value? For sake of simplicity in config, I don't think this is an opt-in.

lukew-cogapp:
Reworked to match, thanks.

Label now escalates body -> annotation -> canvas. Each level takes an array, since a viewport can show several resources at once (paged opening, Scroll, Choice bodies): two or three labels join as `Left, Right`, past that it counts (`5 canvases`). A level is only used when every resource in it is labelled, otherwise the viewport gets named after whichever one happened to have a label.

No opt-in, no config. Nothing described anywhere gives a generic `Image`, so the name is never empty.

Worth flagging: annotation labels are read from the vault here rather than through `getPaintingResource`, which returns bodies only and has a dozen callers. Changing its signature felt out of scope, but say the word if you'd rather it carried the annotation through.

PR #352, opened 2026-08-04
Improve keyboard and motion accessibility of the image viewer

lukew-cogapp:
Three fixes.

1. Exiting OpenSeadragon's full-page mode drops focus, because OSD moves the viewer in and out of the DOM and never restores it (its `setFullPage` saves `previousBody` but never calls `.focus()`). Clover now restores focus to the fullscreen button. OSD's `full-page` event fires on both enter and exit with `{ fullPage: boolean }`, so the handler only acts on exit.

2. OSD eases every zoom and pan. The config now sets `animationTime: 0` when `prefers-reduced-motion: reduce` matches, guarded for SSR and still overridable by an explicit `options.openSeadragon.animationTime`.

3. OSD makes its canvas a tab stop but Clover only styled focus on its own buttons, so keyboard users tabbing to the viewer got no visible indication. There's now a default focus outline on the canvas.

On verification: the canvas focus outline and the accessible name were confirmed in a real browser (2px accent outline on keyboard focus). The fullscreen focus restore is covered by unit tests but was not confirmed in a real browser, because headless Chrome would not enter full-page mode. Flagging that plainly rather than implying it was checked end to end.

Tests were added for each fix, and each was checked to fail when the fix is reverted.


mathewjordan:
Thank you so much for this one @lukew-cogapp !

mathewjordan (approved):

PR #353, opened 2026-08-04
Render the active canvas summary in the About panel

lukew-cogapp:
Clover renders the manifest's `summary` in the About panel but nothing per canvas, so manifests that describe each image individually have that text dropped. IIIF Presentation 3 permits `summary` on a Canvas ("may have", clients "should render").

This change adds the active canvas's summary above the manifest's own in `About.tsx`, reading it live from the vault so it follows the active canvas as you page through the object.

It uses the existing `parent="canvas"` prop on the `Summary` properties component, which was already implemented (giving `id="iiif-canvas-summary"`) but had no callers. Worth mentioning: this suggests the feature was anticipated but never wired up.

It's opt-in via `options.informationPanel.renderCanvasSummary`, defaulting to `false`, because it adds text to a panel consumers already ship.

Verified in a browser against a real manifest with per-canvas summaries: the caption renders and updates as you page between canvases.

Two tests cover it (default omits it, opt-in renders it), plus a docs table row and a short docs section.

### DIFF
--- a/pages/docs/viewer.mdx
+++ b/pages/docs/viewer.mdx
@@ -546,8 +546,14 @@ The information panel is a collapsible panel that displays information about the
 | `options.informationPanel.renderSupplementing` | `boolean`                       | No       | true    |
 | `options.informationPanel.renderToggle`        | `boolean`                       | No       | true    |
 | `options.informationPanel.renderContentSearch` | `boolean`                       | No       | true    |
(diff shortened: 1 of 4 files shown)


mathewjordan:
This has been need for a very long time. Thanks @lukew-cogapp !

mathewjordan (approved):

PR #357, opened 2026-08-24
fix(viewer): only render caption resources as <track> elements

nakamura196:
# Only render caption resources as `<track>` elements

## The problem

An `AnnotationPage` on an A/V canvas may legitimately carry **descriptive**
annotations — the IIIF Presentation API places no restriction on the motivation
or body type of annotations attached to a canvas, and `describing` /
`commenting` bodies on time-based canvases are a normal way to publish a
shot-level catalogue.

`Player.tsx` pushes *every* annotation body into a `<track>` element:

```tsx
annotationNormalized.body.forEach((body) => {
  const annotationBody = vault.get(body.id) as LabeledIIIFExternalWebResource;
  annotationBodies.push(annotationBody);       // ← no filtering
});
…
<Track resource={body} key={body.id} />        // ← <track src={body.id}>
```

This has two consequences.

**1. The browser is asked to fetch subtitle files that do not exist.**
An embedded body (`TextualBody` and friends) holds its text inline and has no
dereferenceable URL, so `<track src="vault://57a8c405">` is a request that can
never succeed.

**2. Duplicate React keys.**
Embedded bodies have no `id` of their own, so the Vault mints a content-derived
`vault://<hash>`. Two bodies with the same text share a hash, so `key={body.id}`
collides and React logs a duplicate-key warning **for each collision, on every
render**.

At realistic catalogue sizes this is not a cosmetic problem. With the browser
console open, the warning volume is enough to make the page unresponsive.

## Reproduction

Manifest: a 27-second video canvas with 8 descriptive annotations. Bodies are
`TextualBody`, targets are time ranges on the canvas.

```jsonc
{
  "type": "Manifest",
  "items": [{
    "id": "https://example.org/practice/canvas/1",
    "type": "Canvas",
    "height": 720, "width": 1280, "duration": 27.0,
    "items": [ /* painting annotation for the video */ ],
    "annotations": [{
      "id": "https://example.org/practice/canvas/1/annos/1",
      "type": "AnnotationPage",
      "items": [{
        "id": "https://example.org/practice/annotation/frame-0000",
        "type": "Annotation",
        "motivation": "describing",
        "body": [
          { "type": "TextualBody", "purpose": "describing",
            "value": "A title card on a black ground.", "language": "en" },
          { "type": "TextualBody", "purpose": "classifying",
            "value": "title card", "language": "en" }
        ],
        "target": "https://example.org/practice/canvas/1#t=0.0,4.0"
      }
      /* … 7 more … */ ]
    }]
  }]
}
```

Open it in `Viewer`. Measured on `main` (`9b0df8a`, v3.14.0):

| annotations on the canvas | duplicate-key warnings on load | `<track>` elements created |
| ---: | ---: | ---: |
| 8 | 14 | 27 |
| 46 | 140 | 247 |
| 217 | **890** | **1222** |

Removing the `annotations` array from the same manifest brings the warning count
to **0**, which isolates the cause to this code path. Making every annotation id
an absolute URI does **not** help — the colliding ids belong to the *bodies*,
not the annotations.

## The change

A new helper, `isCaptionResource()` in `src/lib/annotation-helpers.ts`, filters
annotation bodies down to resources a `<track>` can actually fetch:

- rejects bodies with no `id`
- rejects Vault-minted `vault://<hash>` ids
- rejects `TextualBody`
- accepts a declared `format` of `text/vtt` or `application/x-subrip`
- with no declared `format`, falls back to a `.vtt` / `.srt` file extension

The remaining keys are made unique by including the annotation page id and the
index, so a manifest that legitimately lists the same caption resource twice
cannot collide either.

After the change, all three manifests above produce **0 warnings and 0 stray
`<track>` elements**, while the information panel still lists every annotation
(8 / 46 / 217 rows).

## Not changed

Manifests that declare real caption resources behave exactly as before. The
`ignoreCaptionLabels` option, the WebVTT cue rendering and the
`PointSelector` → `text/vtt` inference are all untouched.

## Verification

Ran the three `pull_request` CI lanes locally (macOS, Node 25):

- `npm run test:ci` — **78 files, 348 passed, 1 skipped**. Includes 7 new cases
  for `isCaptionResource`, covering declared formats, extension fallback,
  `TextualBody`, Vault-minted ids, missing ids and non-caption formats.
- `npm run build` — clean, 1407 modules transformed.
- `npx playwright test e2e/wc.spec.ts` (from `playwright/`, as `wc-e2e.yml`
  does) — 1 passed.
- `npm run typecheck` — reports the same 7 pre-existing errors as `main` (all in
  `src/components/Slider/Items/Items.test.tsx`), none in the touched files.

Not run locally: the React 18/19 example builds in `compat.yml` beyond the
shared library build. The change is a single predicate with no React API
surface, so I would not expect a version-specific effect, but I have not
verified that lane.

Manual: verified in a Next.js app against the three manifests above, counting
duplicate-key warnings and `<track>` elements in the browser before and after.
Warnings went 14 / 140 / 890 → **0 / 0 / 0**; `<track>` elements went
27 / 247 / 1222 → **0 / 0 / 0**; the information panel still listed all
8 / 46 / 217 annotations.

## Context

Found while building a shot-level catalogue of amateur films for a Japanese
research project (JSPS KAKENHI 24K22434). We publish descriptive annotations on
video canvases and embed Clover to demonstrate that the output is readable by a
third-party viewer, so this path is exercised heavily.

Two related gaps are **not** addressed here, to keep this change small. Happy to
open separate issues if they would be welcome:

1. **Spatial annotations on A/V canvases are ignored.** `#xywh` is valid on a
   time-based canvas per the Presentation API, but overlay rendering goes
   through `openSeadragonViewer.getOverlayById()`, which is `null` for A/V. The
   documentation states that spatial overlays are image-only; the underlying
   data is expressible, so a renderer for A/V canvases would be useful.
2. **Time-based annotations are not synchronised to the player.** The only path
   that seeks the player is WebVTT cue playback, reachable via an external
   `.vtt` file or a `PointSelector`. An annotation with
   `target: canvas#t=A,B` appears in the list but is not linked to the
   timeline, and `PointSelector` cannot express the end of a range.

### DIFF
--- a/src/components/Viewer/Player/Player.tsx
+++ b/src/components/Viewer/Player/Player.tsx
@@ -12,6 +12,7 @@ import AudioVisualizer from "src/components/Viewer/Player/AudioVisualizer";
 import { LabeledIIIFExternalWebResource } from "src/types/presentation-3";
 import { PlayerWrapper } from "src/components/Viewer/Player/Player.styled";
 import Track from "src/components/Viewer/Player/Track";
+import { isCaptionResource } from "src/lib/annotation-helpers";
 import { getPaintingResource } from "src/hooks/use-iiif";
 import { isHls } from "src/lib/hls";
 
@@ -333,16 +334,32 @@ const Player: React.FC<PlayerProps> = ({
                 const annotationBody = vault.get(
                   body.id,
                 ) as LabeledIIIFExternalWebResource;
-                annotationBodies.push(annotationBody);
+
+                /**
+                 * Only external caption resources belong in a <track> element.
+                 *
+                 * An AnnotationPage on an A/V canvas may legitimately carry
+                 * descriptive annotations whose bodies are embedded TextualBody
+                 * resources (no `id`, no dereferenceable URL). Rendering those
+                 * as <track src="..."> makes the browser request a subtitle
+                 * file that does not exist.
+                 *
+                 * Embedded bodies also have no `id` of their own, so the Vault
+                 * mints a content-derived `vault://<hash>`. Two bodies with the
+                 * same text share a hash, which produced duplicate React keys
+                 * (one warning per collision, per render).
+                 */
+                if (isCaptionResource(annotationBody))
+                  annotationBodies.push(annotationBody);
               });
             });
 
-            return annotationBodies.map((body) => {
+            return annotationBodies.map((body, index) => {
               return (
                 <Track
                   resource={body}
                   ignoreCaptionLabels={configOptions.ignoreCaptionLabels || []}
-                  key={body.id}
+                  key={`${annotationPage.id}-${body.id}-${index}`}
                 />
               );
             });
--- a/src/lib/annotation-helpers.test.ts
+++ b/src/lib/annotation-helpers.test.ts
@@ -2,6 +2,7 @@ import {
   getLanguageDirection,
   parseAnnotationTarget,
   filterAnnotationsByMotivation,
+  isCaptionResource,
   AnnotationTargetExtended,
 } from "./annotation-helpers";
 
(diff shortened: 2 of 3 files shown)


mathewjordan (approved):
Thank you Satoru! This has bothered me for a long time. Best to you.

nakamura196:
Thank you for the quick review, Mathew — much appreciated.

Before this lands I put the change through a harder self-review, and it turned up two cases where a caption that used to render would have stopped rendering. Both are spec-valid manifests, so I've pushed a second commit rather than let them ship. Sorry for the churn after your approval — please take another look when you have a moment.

**1. `format: "text/vtt; charset=utf-8"` was dropped.** The Presentation API says `format` "should be the value of the Content-Type header returned when the resource is dereferenced", and that header routinely carries a charset parameter. I was comparing the raw string. Parameters are now stripped first.

**2. A body with no `format` whose id has no file extension was dropped.** `format` is a SHOULD, not a MUST, and captions served from an extensionless URL (`.../master_files/<id>/captions`) are common. Such a body is now kept when its id is dereferenceable.

The predicate now rejects what it can show is *not* a caption rather than accepting only what it recognises — dropping a real caption is worse than keeping a dead menu entry. A recognisable non-caption extension still settles it, so that allowance doesn't swallow every other file type. `text/webvtt` and `text/srt` are accepted too; both are published in the wild.

I also dropped `index` from the React key. With it, inserting a caption ahead of an existing one remounts the rest, and remounting a `<track>` resets `TextTrack.mode` — a viewer's captions would silently switch off. The page id and body id identify the track on their own.

**On testing.** You were approving a change to a branch with no coverage: every existing `Player` test passes `annotationResources: []`, so none of them reached the `<track>` code at all. That's now covered — five tests render a video canvas whose AnnotationPage carries a mix of bodies and assert which ones become tracks. 363 tests pass (was 348); build and the WC e2e both pass.

**Two findings from checking public manifests**, in case they're useful:

- Wellcome Collection's `b16659090` publishes a **PDF transcript** as a `supplementing` body on a video canvas, so the browser was being handed a PDF to parse as WebVTT. That's now covered by a test.
- Cookbook recipe **0074 (multiple-language captions)** wraps two WebVTT files in a `Choice`, which carries no id of its own. It wasn't rendered before this change either — the Vault-minted id was unfetchable — so this isn't a regression, but expanding `Choice.items` would make that recipe work for the first time. I've left it out to keep this PR to one thing. Happy to add it here or open a follow-up, whichever you prefer.

One caveat I should state plainly: the duplicate-key warning counts in the description come from our own manifests. I checked the Cookbook and a number of Northwestern and Wellcome manifests and none of them reproduce it, because each A/V canvas there carries at most one annotation body. The new key is defensive rather than demonstrable on public data.

nakamura196:
Our messages crossed — this merged a few minutes before I pushed the corrections above, so they are not on `main`. I've opened #358 with them. The two dropped-caption cases are live on `main` right now, so that one is worth a look when you have a moment.

PR #358, opened 2026-08-24
fix(viewer): keep captions #357 dropped, and cover the track branch

nakamura196:
Follow-up to #357. I pushed these corrections to that branch a few minutes after it merged, so they missed the merge — opening them here instead. Sorry for the split.

Self-review of #357 found two cases where a caption that used to render stopped rendering. Both are spec-valid manifests, and both are on `main` right now.

## 1. `format: "text/vtt; charset=utf-8"` is dropped

The Presentation API says `format` "should be the value of the Content-Type header returned when the resource is dereferenced", and that header routinely carries a charset parameter. #357 compared the raw string:

```ts
if (format) return format === "text/vtt" || format === "application/x-subrip";
```

So a manifest that follows the spec's own recommendation loses its captions. Parameters are now stripped before comparing.

## 2. A body with no `format` and no file extension is dropped

`format` is a SHOULD, not a MUST. #357 fell back to a `.vtt` / `.srt` extension test, which drops captions served from an extensionless URL — `.../master_files/<id>/captions` and similar are common.

## The shape of the fix

The predicate now rejects what it can show is *not* a caption, rather than accepting only what it recognises. Dropping a real caption is worse than keeping a dead menu entry, so anything ambiguous is kept:

- no `id`, a `vault://` id, or `TextualBody` — rejected, as before
- a declared `format` — accepted only if it is a caption format (`text/vtt`, `text/webvtt`, `text/srt`, `application/x-subrip`)
- no declared `format` — a caption extension accepts, a recognisable non-caption extension rejects, and an extensionless dereferenceable id is kept

That last rule is what restores case 2 without letting every other file type through.

## Also: `index` removed from the React key

#357 keyed tracks as `${annotationPage.id}-${body.id}-${index}`. With the index in the key, inserting a caption ahead of an existing one remounts the rest, and remounting a `<track>` resets `TextTrack.mode` — a viewer's captions silently switch off. The page id and body id identify the track on their own. There is no path in Clover today that reorders that array, so this is defensive.

## Tests

`Player`'s `<track>` branch had no coverage at all: every existing Player test passes `annotationResources: []`, so none of them reached the code #357 changed. Five tests now render a video canvas whose AnnotationPage carries a mix of bodies and assert which ones become tracks — including the embedded-body case the filter exists for, and a PDF transcript published as a `supplementing` body, taken from a real Wellcome Collection manifest (`b16659090`). Ten more cover the predicate's edge cases directly.

363 tests pass (was 348 after #357). Build and the Web Components e2e both pass; I ran all three `pull_request` workflows locally.

## Not addressed

Cookbook recipe **0074 (multiple-language captions)** wraps two WebVTT files in a `Choice`, which has no id of its own, so it is rejected. It was not rendered before #357 either — the Vault-minted id was unfetchable — so this is not a regression, but expanding `Choice.items` would make that recipe work for the first time. Left out to keep this to one thing; happy to open it separately if you'd like it.

### DIFF
--- a/src/components/Viewer/Player/Player.test.tsx
+++ b/src/components/Viewer/Player/Player.test.tsx
@@ -195,4 +195,176 @@ describe("Player component", () => {
     // Test for the audio visualizer
     expect(screen.getByRole("presentation")).toBeInTheDocument();
   });
+
+  /* The `<track>` branch had no coverage: every existing test passes
+     `annotationResources: []`, so none of them reach it. These render a video
+     canvas whose AnnotationPage carries a mix of bodies and assert which ones
+     become tracks. */
+  describe("caption tracks", () => {
+    const CANVAS = "https://example.org/canvas/1";
+
+    function renderWithBodies(bodies: any[]) {
+      const painting = {
+        id: "https://example.org/video.mp4",
+        type: "Video",
+        format: "video/mp4",
+        height: 720,
+        width: 1280,
+        duration: 27,
+      };
+
+      const annotationResources = [
+        {
+          id: "https://example.org/canvas/1/annotations/1",
+          type: "AnnotationPage",
+          items: [
+            {
+              id: "https://example.org/annotation/1",
+              type: "Annotation",
+              motivation: ["supplementing"],
+              body: bodies,
+              target: CANVAS,
+            },
+          ],
+        },
+      ];
+
+      const vault = new Vault();
+      vault.loadSync("", {
+        "@context": "http://iiif.io/api/presentation/3/context.json",
+        id: "https://example.org/manifest.json",
+        type: "Manifest",
+        label: { none: ["Captions"] },
+        items: [
+          {
+            id: CANVAS,
+            type: "Canvas",
+            height: 720,
+            width: 1280,
+            duration: 27,
+            items: [
+              {
+                id: "https://example.org/canvas/1/page/1",
+                type: "AnnotationPage",
+                items: [
+                  {
+                    id: "https://example.org/canvas/1/page/1/annotation/1",
+                    type: "Annotation",
+                    motivation: "painting",
+                    body: painting,
+                    target: CANVAS,
+                  },
+                ],
+              },
+            ],
+            annotations: annotationResources,
+          },
+        ],
+      } as any);
+
+      const { container } = render(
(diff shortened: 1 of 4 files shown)


nakamura196:
One more commit, narrowing rather than widening — and it corrects something I got wrong above.

I had `text/srt` and `application/x-subrip` in the accepted formats. **A `<track>` renders WebVTT and nothing else.** The HTML spec allows no other format and browsers ignore one silently, with no fallback, so a SubRip body produces a caption menu entry that shows nothing when a viewer picks it.

Publishers do send SubRip — but as `supplementing` transcripts for a panel to render, not as captions. Measured against 400 Indiana Avalon manifests: every body under `/captions` is `text/vtt`, while `text/srt` and `application/x-subrip` appear only under `/transcripts`. Avalon's own code agrees; it validates captions as vtt/srt/subrip and rewrites SubRip to `text/vtt` before it reaches the manifest (`supplemental_file.rb`, `iiif_canvas_presenter.rb`).

Counting `<track>` elements against those 400 manifests:

| | rendered | of them working |
|---|---:|---:|
| `main` before #357 | 1378 | 54 |
| #357 as merged | 363 | 54 |
| this PR, before this commit | 705 | 54 |
| this PR now | **54** | **54** |

No manifest in the corpus loses a caption that used to work, at any step. What goes away is dead menu entries — and #357 as merged still leaves 309 of them, all from `application/x-subrip`.

`text/webvtt` stays. It is a non-standard spelling of a format the browser can actually render.

362 tests pass; build and the WC e2e pass.

---

On **`Choice`**: this is now the only real gap I can find in public data. Cookbook recipe 0074 wraps two WebVTT files in a `Choice` with no id of its own, and the National Library of Wales publishes the same structure. Neither renders today and neither rendered before #357, so it is not a regression — but the Vault does return both VTT resources when you expand `Choice.items`, so it looks like a small change. Say the word and I'll open it separately; I've kept it out of here to keep this to one thing.

One caveat on my own numbers, stated plainly: the corpus above is Indiana Avalon plus Northwestern, Wellcome, the Cookbook and the Internet Archive. Yale, Texas A&M, Cornell, TCD and a few others blocked automated fetches, so they are not represented.

mathewjordan (approved):

PR #362, opened 2026-09-04
Fix WCAG 2.2 AA failures found in a viewer audit

lukew-cogapp:
We had the viewer demo audited against WCAG 2.2 AA by an accessibility specialist, and these are the code-level failures that came out of it. The audit ran against the published demo, so a few of its findings were already fixed by the vanilla CSS work in #361. Those are noted at the bottom rather than fixed again.

Six fixes, each its own commit.

## Accessible names

The button over a canvas was called "Open" or "Close". It swaps the static placeholder for the zoomable viewer, so it now says that. The rail's controls said "Next", "Previous" and "Search" without saying what of, and now say "Next item", "Search items" and so on.

These use new keys (`imageViewerOpen`, `sliderNext`, …) rather than repointing `commonOpen` and friends, because those carry translations in seven locales that would have kept the old vague wording. The old keys still ship but nothing reads them; they are listed under Deprecated in the changelog. All seven locales have the new strings, so nobody falls back to English.

## Duplicate ids on the icons

The toggle and the map's zoom controls named their icons with `aria-labelledby` pointing at a hard-coded id. Two viewers on one page emit that id twice and every button resolves to the first title. They use `role="img"` with a bare `<title>` now, which needs no id.

Worth flagging: I first tried `useId` here, then found it does not help. Two `<clover-viewer>` elements are two React roots and `useId` only counts within one. I then tried deriving the id from `viewerId`, which does not help either: `createDefaultState()` calls `uuidv4()` once at module load, and `Viewer/index.tsx` spreads that frozen state into every provider, so the `?? uuidv4()` fallback never runs and every viewer on a page shares one id. That looks like a bug worth its own issue; it is not fixed here.

## `role="group"` inside `<dl>`

Metadata pairs were wrapped in `<div role="group">`. A `dl` accepts a plain `div` between itself and its `dt`/`dd` children, but not that role, and it broke the pairs' association. The `data-label` hook is unchanged, so anything selecting on it still works.

## Controls escaping the canvas when zoomed

At 400% zoom on a 1280px viewport (the reflow condition in 1.4.10) the control cluster ran off the bottom and the last buttons could not be reached. It is absolutely positioned with no bound of its own.

The fix is a `--top` custom property carrying whatever offset the compound rules set (they push it as far as `6.236rem` to clear the toggles), so `max-height` can be computed against it. A plain `max-height: calc(100% - 2rem)` does not work, because it assumes `top: 1rem`.

Wrapping also widens the box to whatever space it is offered, so the row needs an explicit `justify-content: flex-end` or it packs itself against the far edge and lands under the navigator.

There is a Playwright spec for this in `playwright/e2e/reflow.spec.ts`, added to the existing `wc-e2e` job. It drives the real web component against a local manifest, so it adds no network dependency. Each assertion was checked against a deliberately broken stylesheet to confirm it fails.

## Focus and announcements

Radix already gives the information panel's tab panel a tab stop, so its content could be scrolled from the keyboard, but nothing showed where that focus had landed. That is a focus ring, nothing more.

Changing canvas replaces the image in place, which a screen reader has no reason to report, so stepping through the rail with the next button was silent. A polite live region names the new canvas, falling back to "Item 2 of 4" where the canvas has no label, which is common. This is the first key with interpolation, and `useCloverTranslation`'s English fallback did not interpolate, so that is fixed too.

## Checks

`npm test` (426 passing), `npm run typecheck` (5 errors, all pre-existing in files this branch does not touch), `npm run build`, and prettier are all clean. The new e2e specs pass. The layout fix was measured across 30 combinations of canvas height and control configuration at three viewport widths.

One thing I left alone: `Toggle` still takes an `isMedia` prop that `Painting` can never pass as true, and `Toggle.css` has rules for it describing a centred play affordance. Removing the prop and those styles is a design call rather than an accessibility one, so it seemed out of scope here.


mathewjordan (approved):
✅

PR #363, opened 2026-09-10
Make information panel visibility and tab selection deterministic

bmquinn:
## Summary

Clarifies information panel behavior with named visibility and tab-selection logic.

- Every `render*` option (`renderAbout`, `renderAnnotation`, `renderContents`, `renderContentSearch`, `renderToggle`) defaults to on; a tab is only withheld when the option is explicitly `false`.
- Annotations (including VTT chapters and transcripts), a Contents table built from Manifest `structures`, and an IIIF Content Search service are treated as primary content. When no toggle is rendered, any of them makes the panel visible even if `open` or `renderAbout` is false.
- When Clover renders its toggle, the toggle owns open and closed: `open` sets the initial state and the reader controls it from there. Primary content no longer forces the panel open in that case, so the toggle can always close it.
- Tab selection falls back to the first available tab while resources load, and applies `defaultTab` once that tab's data is present.

## Documentation

Updates the viewer documentation to clarify:
- that `render*` options default to on and only an explicit `false` hides a tab;
- how `open`, `renderToggle`, About, annotations, Contents, and Content Search interact to make the panel visible;
- how `defaultTab` and fallback selection work;
- the correct shared Annotations tab identifier;
- which legacy information-panel options no longer control tab rendering.

## Testing

- Added coverage for panel visibility with About disabled or closed, for annotations, Content Search, and Manifest `structures`.
- Added coverage that a rendered toggle keeps the panel closed until the reader opens it.
- Added coverage that annotations render when no `render*` options are set.
- Added coverage for annotation and Content Search tab availability and selection.

Confirmed functionality in a `dc-nextjs` application running `clover-iiif` on this branch:
SCR-20260910-mycn

Example with VTT content on current `v3.16.0`:
current

Same work on this branch:
new_logic


mathewjordan (approved):
🚀

